PR #17, opened 2024-06-05
Feat/16 get all errors in form

michalina-majewska:


lib/src/form_group_cubit/form_group_cubit.dart
```diff
@@ -344,6 +344,20 @@ class FormGroupState with EquatableMixin {
   /// Returns true if fields are currently being validated.
   final bool validating;
 
+  /// List of this form's fields including subforms' fields.
+  Iterable<FieldCubit<dynamic, dynamic>> get allFields {
+    return [
+      ...fields,
+      for (final subform in subforms) ...subform.state.allFields,
+    ];
```
shilangyu, line 352: What I meant is to actually return an Iterable to not allocate a list

```suggestion
    return fields.followedBy(subforms.expand((e) => e.state.allFields));
```

PR #19, opened 2024-06-12
Use clear() in the example

michalina-majewska:
Add 'set to initial' buttons
(It seemed a bit complicated,  fields' initialValues could be public)

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,8 @@ The package contains a collection of field cubits useful for implementing common
 - `SingleSelectFieldCubit` - specialization of `FieldCubit` for a single choice of value from list of options,
 - `MultiSelectFieldCubit` - specialization of `FieldCubit` for a multiple choice of values from list of options.
 
+`TextFieldCubit`, `SingleSelectFieldCubit`, `MultiSelectFieldCubit` contain the `clear()` method that resets the value of the field to the initial value by calling `reset()`. You can also call `reset()` as it is defined in the `FieldCubit` class.
+
 ### Creating custom `FieldCubit`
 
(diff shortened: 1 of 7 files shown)


example/lib/widgets/app_text_field.dart
```diff
@@ -59,19 +61,42 @@ class AppTextField extends HookWidget {
...
+          ),
+        ),
+        const SizedBox(width: 16),
+        ElevatedButton(
+          onPressed: () {
+            {
+              textEditingController.clear();
+              setValue('');
+            }
+          },
```
shilangyu, line 89:
```suggestion
          onPressed: () {
            textEditingController.clear();
            setValue('');
          },
```

README.md
```diff
@@ -213,6 +213,8 @@ The package contains a collection of field cubits useful for implementing common
 - `SingleSelectFieldCubit` - specialization of `FieldCubit` for a single choice of value from list of options,
 - `MultiSelectFieldCubit` - specialization of `FieldCubit` for a multiple choice of values from list of options.
 
+`TextFieldCubit`, `SingleSelectFieldCubit`, `MultiSelectFieldCubit` contain the `clear()` method that resets the value of the field to the initial value by calling `reset()`. You can also call `reset()` as it is defined in the `FieldCubit` class.
```
shilangyu, line 216:
```suggestion
`TextFieldCubit`, `SingleSelectFieldCubit`, and `MultiSelectFieldCubit` contain the `clear()` method that resets the value of the field to the initial value by calling `reset()`. You can also call `reset()` as it is defined in the `FieldCubit` class.
```

shilangyu (approved):

PR #22, opened 2025-01-02
upgrade rxdart version to 0.28.0

papeye:
Changes required to upgrade flutter version in [template app](https://github.com/leancodepl/leancode_flutter_template), specifically upgrade of rxdart to 0.28.0 is needed 

Please let me know about the approach taken here. For now in [pr upgrading flutter in template](https://github.com/leancodepl/leancode_flutter_template/pull/340) i'm using ref from the last commit in this branch as dependency; previously it was imported as package. Should there be new release planned or should we stick to getting package from ref? If so am I allowed to merge it to main?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+
+
+## 0.1.1
+
+* Bumped `rxdart` to `^0.28.0`.
+
 ## 0.1.0
 
 * Write README.md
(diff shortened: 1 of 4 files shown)


piotruela (changes requested):
Please also update changelog and package version in `pubspec.yaml`

piotruela (approved):

PR #42, opened 2026-06-19
Refactor/cubit to notifier migration

aydinguven-leancode:
[LMG-394] 

- Migration from Cubit to a Value Notifier based state management
- Dropped dependencies: `flutter_bloc`, `rx_dart`, flutter_hooks`, `equatable`
- Name changes for relevant classes
- Relevant changes on tests and example project 

[LMG-404] 
- Partially modified Readme.md and Changelog.md files about the above migration
- Introduced `Migration.md` file 
- Added a few more examples as well as short descriptions to each example page

[LMG-394]: https://leancode.atlassian.net/browse/LMG-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LMG-404]: https://leancode.atlassian.net/browse/LMG-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pubspec.yaml
```diff
@@ -9,14 +9,10 @@ environment:
...
     sdk: flutter
-  flutter_bloc: ^9.0.0
-  rxdart: ^0.28.0
 
 dev_dependencies:
-  bloc_test: ^10.0.0
   flutter_test:
     sdk: flutter
-  leancode_lint: ^15.0.0
+  leancode_lint: ^15.0.0
```
PiotrRogulski: Keep final newline

lib/src/field/field_controller.dart
```diff
@@ -0,0 +1,373 @@
...
+/// Translates an error to a string.
+typedef ErrorTranslator<E extends Object> = String Function(E);
+
+/// A single form field which can be validated.
+/// Stores the current value, error text, and whether autovalidate is on.
+/// [T] is the held value, [E] is the type of an error. [E] cannot be nullable
+/// to be able to unambiguously detect lack of errors.
+///
+/// If autovalidate is true, the validator will be run after each field change.
+// ignore_for_file: avoid_positional_boolean_parameters
```
PiotrRogulski: This package depends on `leancode_lint 15`, which has this lint disabled (in version 14) – no need to ignore it, I think

lib/src/form_group/form_group_controller.dart
```diff
@@ -0,0 +1,389 @@
...
+/// A form is a tree which can be recursively defined:
+///   1. A form is the root of its own form tree
+///   2. A form has direct leaves, which are fields
+///   3. A form can have subtrees, which are forms (called subforms)
+///
+/// Most methods broadcast to all subforms.
+///
+/// Introducing cycles in forms is not supported and not checked against (most
+/// likely will cause a stack overflow somewhere).
+// ignore_for_file: avoid_positional_boolean_parameters
```
PiotrRogulski: Ditto

lib/src/field/field_controller.dart
```diff
@@ -0,0 +1,373 @@
...
+class FieldController<T, E extends Object>
+    extends ValueNotifier<FieldState<T, E>> {
```
PiotrRogulski: Do we want to extend `ValueNotifier` directly, instead of e.g. `ValueListenable` and adding our own value field? `ValueNotifier` exposes a public setter, so that anyone can create and set arbitrary `FieldState` values

lib/src/field/field_controller.dart
```diff
@@ -0,0 +1,373 @@
...
+
+  Timer? _debounceTimer;
+
+  CancelableFuture<E?>? _asyncValidationFuture;
+
+  VoidCallback? _fieldsSubscriptionCleanup;
+
+  /// The current state. Alias for [value] kept for readability at call sites
+  /// that previously read `cubit.state`.
+  FieldState<T, E> get state => value;
```
PiotrRogulski: I wouldn't do that tbh. The entire undertaking is extremely breaking anyway, so I don't see a value in keeping a bloc-related alias

example/lib/screens/delivery_form.dart
```diff
@@ -24,27 +25,38 @@ class DeliveryListForm extends StatelessWidget {
...
+              bold('cascades'),
+              plain(' to every subform.'),
+            ]),
+            ...controller.deliveryList.map(
+              (e) => ConsumerSubform(
+                key: ValueKey(e.hashCode),
+                form: e,
+                onRemove: controller.removeConsumer,
+              ),
+            ),
```
PiotrRogulski: Use a `for` loop

example/lib/screens/optimized_rendering_form.dart
```diff
@@ -0,0 +1,262 @@
...
+class _OptimizedRenderingForm extends StatelessWidget {
+  const _OptimizedRenderingForm();
+
+  @override
+  Widget build(BuildContext context) {
+    final controller = context.read<OptimizedRenderingFormController>();
+    return FormPage(
+      title: 'Optimized Rendering',
+      child: SingleChildScrollView(
+        child: Column(
```
PiotrRogulski: `SingleChildScrollView` + `Column` -> `ListView`?

example/lib/controllers/focusable_text_field_controller.dart
```diff
@@ -0,0 +1,26 @@
...
+  FocusableTextFieldController({
+    super.initialValue,
+    super.validator,
+    super.asyncValidator,
+    super.asyncValidationDebounce,
+    super.name,
+  });
+
+  /// The focus node of the field.
+  final focusNode = FocusNode();
```
Albert221: Use `debugLabel`

example/lib/controllers/focusable_text_field_controller.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:flutter/widgets.dart';
+import 'package:leancode_forms/leancode_forms.dart';
+
+/// A [TextFieldController] that owns a [FocusNode] for scroll-to-error flows.
+class FocusableTextFieldController<E extends Object>
```
Albert221: All form controls are usually focusable, I don't think having a `Focusable...` wrapper for each of the `...FieldController` is a good approach

Albert221: Wait why is it in example and not in the lib itself?

EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
+# Examples
+
+Side-by-side snippets showing the same form scenario in three flavors:
+
+1. **0.1.x (pre-migration)** — built on `flutter_bloc` + `rxdart`
```
Albert221: I don't care about older versions in EXAMPLE. That's what changelog, git history and migration document is for

EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+2. **0.2.0 with `FieldBuilder`** — the recommended default; tiny wrapper around `ValueListenableBuilder`
+3. **0.2.0 with `ValueListenableBuilder`** directly — for cases where you want the SDK primitive (e.g. the `child:` optimization)
```
Albert221: What's the use-case for having this approach? Looking at just the API I think it would be better and simpler to keep one API for the builders

example/pubspec.yaml
```diff
@@ -1,26 +1,22 @@
...
     path: ..
-  rxdart: ^0.28.0
+  provider: ^6.1.2
 
 dev_dependencies:
-  bloc_test: ^10.0.0
   flutter_test:
     sdk: flutter
-  leancode_lint: ^15.0.0
+  leancode_lint: ^6.0.0
```
PiotrRogulski: Why such a downgrade?

EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+    await Future<void>.delayed(const Duration(milliseconds: 700));
+    return taken.contains(value) ? ValidationError.emailTaken : null;
+  }
+}
+
+// Render:
+FieldBuilder<String, ValidationError>(
+  field: context.read<SignupController>().email,
+  builder: (context, state) => TextFormField(
+    controller: context.read<SignupController>().email.textController,
```
Albert221: We already passed the field controller, why fetch it again? Please just add the field controller itself into the builder callback params.

EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+  Future<ValidationError?> _checkEmail(String value) async {
+    final taken = ['taken@example.com'];
+    await Future<void>.delayed(const Duration(milliseconds: 700));
+    return taken.contains(value) ? ValidationError.emailTaken : null;
+  }
+}
+
+// Render:
+FieldBuilder<String, ValidationError>(
+  field: context.read<SignupController>().email,
```
Albert221: Using `context.read` instead of `watch/select` in `build()` method is an antipattern

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+> Upgrading from 0.1.x? See [MIGRATION.md](./MIGRATION.md) for a step-by-step guide.
+
+* **Breaking:** We've rebuilt the library on `ValueNotifier` / `ChangeNotifier`. That means `flutter_bloc` and `rxdart` no longer are a dependency.
+* **Breaking:** Renamed core classes:
+  * `FieldCubit` → `FieldController`
+  * `TextFieldCubit` → `TextFieldController`
+  * `BooleanFieldCubit` → `BooleanFieldController`
+  * `SingleSelectFieldCubit` → `SingleSelectFieldController`
+  * `MultiSelectFieldCubit` → `MultiSelectFieldController`
+  * `FormGroupCubit` → `FormGroupController`
```
Albert221: I'm not a fan of the `...Controller` naming. It's too close to `TextEditingController`. I'd suggest maybe `Form...Field` like `FormBooleanField` or `FormTextField`

lib/src/form_group/form_group_controller.dart
```diff
@@ -0,0 +1,389 @@
...
+class FormGroupController extends ValueNotifier<FormGroupState> {
```
cupofme: Why does entity called `FormGroupController` manages entities called `FieldController`? Naming looks a bit inconsistent to me

EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+```
+
+Note the `controller: field.textController` — `TextFieldController` owns its own `TextEditingController` now, kept in two-way sync with the field value. No `onChanged: field.setValue` needed; programmatic resets (`field.reset()`, `field.clear()`) propagate to the visible text automatically.
+
+### 2. A form with submit-time validation
+
+```dart
+class SimpleFormController extends FormGroupController {
+  SimpleFormController() {
+    registerFields([firstName, email]);
```
cupofme, line 38: Can it be an abstract getter `fields` in `FormGroupController` that needs to be implemented in subclasses? This way it'll be required and therefore one less thing that can be forgotten.

Something like this:
```dart
class SimpleFormController extends FormGroupController {
  SimpleFormController();
  
  @override
  List<FieldController> get fields => [
    firstName, 
    email,
  ];

...
}
```

PiotrRogulski: Having `registerFields` allows dynamically adding fields as needed (for example, conditional fields). Although for that, an `unregisterFields` method would be needed 🤔

example/test/screens/simple_form_test.dart
```diff
@@ -1,47 +1,46 @@
...
+
+    expect(controller.email.value.value, 'john@email.com');
+  });
+
+  test('sets ValidationError.emailTaken when email is taken', () async {
+    final controller = SimpleFormController();
+    addTearDown(controller.dispose);
+
+    controller.email.setValue('john@email.com');
+    await Future<void>.delayed(const Duration(seconds: 2));
```
PiotrRogulski, line 21: It would be nice to have a way to "await validation"

lib/src/form_group/form_group_controller.dart
```diff
@@ -0,0 +1,389 @@
...
+  // comparison, since FieldController/FormGroupController don't override `==`).
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is FormGroupState &&
+          wasModified == other.wasModified &&
+          const ListEquality<FieldController<dynamic, dynamic>>()
+              .equals(fields, other.fields) &&
+          const SetEquality<FormGroupController>()
+              .equals(subforms, other.subforms) &&
```
PiotrRogulski: Consider using `listEquals` and `setEquals` from Flutter

example/lib/controllers/focusable_text_field_controller.dart
aydinguven-leancode: Yep, I agree with you. Only, here this is a faithful copy of what was already added in the package. I assumed the point was to showcase how focus node could be added.

I could remove this example or maybe a better documentation of the purpose of this, what do you think?                       
                                                                       
Lastly, I see that the "Controller" suffix is causing some confusion here. i saw your comment on that and will answer there.

lib/src/form_group/form_group_controller.dart
aydinguven-leancode: Fixed.

lib/src/field/field_controller.dart
aydinguven-leancode: done. I also brought back the `leancode_lint 15` which was removed unintentionally while trying to run the example project.

example/pubspec.yaml
aydinguven-leancode: Fixed. Removed unintentionally.

example/lib/controllers/focusable_text_field_controller.dart
aydinguven-leancode: Added.

CHANGELOG.md
aydinguven-leancode: Yes,  I agree with you. That `...Controller` will cause confusion. I would like to change it but
I'm not sure whether `...Field` fits either since these are state holder classes, not widgets. 

I am not that creative in this example... I have these weak alternatives, let me know what do you think:
- ...Wrapper/Wrap - almost like a tortilla
- ...Orchestrator - might be too long
- ...Cell
- ...Pip/Blip

EXAMPLES.md
aydinguven-leancode: Yes, I agree on changing it into one API. 

I'll add the `child` property to `FieldBuilder` and forward it so that the consumerw have the control over the builder parameters: `context`, `value`, `child`.

EXAMPLES.md
aydinguven-leancode: Removed the lines and comparison.

EXAMPLES.md
aydinguven-leancode: I removed these bits from the EXAMPLES.md file as a fix for the comment above.

pubspec.yaml
aydinguven-leancode: Fixed

example/test/screens/simple_form_test.dart
aydinguven-leancode: Yes, I agree with that. Just to inform you, there are 3 tasks related to validation: LMG-399, LMG-396, LMG-398. I believe that will be a better place to add this change. 
 
 I would add a TODO here and add the changes when tackling that task.

lib/src/form_group/form_group_controller.dart
aydinguven-leancode: Yup, changed. Thanks for suggestion.

lib/src/field/field_controller.dart
aydinguven-leancode: Makes sense. I would like to avoid `controller.value.value` therefore looked for a different name/alias. 

I am planning to change this into `data` , let me know if you have anything else in mind.

EXAMPLES.md
aydinguven-leancode: Yep, you're right. Fixed by changing into a `select`, also in example codes.

example/lib/screens/delivery_form.dart
aydinguven-leancode: Fixed.

example/lib/screens/optimized_rendering_form.dart
aydinguven-leancode: Fixed

CHANGELOG.md
KamilSztandur: We assumed:
- `...Field` is for UI
- `...FieldController` is for controller for that UI

Example: `PhoneNumberField` (widget) + `PhoneNumberFieldController` (controller for that widget)

aydinguven-leancode: Alright, after debating with Kamil Sztandur we 've decided to keep it as Controller but watch for ambiguity, and decide if this needs renaming into a better fitting name. Closing this one.

lib/src/form_group/form_group_controller.dart
aydinguven-leancode: After debating with Kamil Sztandur, decided to move this into FormController. Closing this one.

lib/src/field/field_controller.dart
aydinguven-leancode: Ok changed into `value` after discussion

EXAMPLES.md
aydinguven-leancode: Hi, this will be a part of a different task/PR.
I am parking the idea and resolving this.

lib/src/field/field_controller.dart
aydinguven-leancode: Changed. We are now using ValueListenable in this class.

lib/src/field/advanced_field_controller.dart
```diff
@@ -0,0 +1,389 @@
...
+/// The state of a [AdvancedFieldController].
+///
+/// Value-equal: two [AdvancedFieldState]s are equal when every field matches. This
+/// is required for [ValueNotifier]'s built-in dedup — without it, every
+/// [AdvancedFieldController.setValue] call would notify listeners even when the new
+/// state is identical to the old one.
+///
+/// **Maintainer note:** when adding a new field below, you MUST also add it
+/// to both [operator ==] and [hashCode] at the bottom of the class, otherwise
+/// the new field is silently invisible to dedup and structural comparisons.
```
PiotrRogulski: This note shouldn't be part of the dartdoc – it shouldn't show up in hover popups or in the generated docs:

Image

(in AdvancedFormState too)

aydinguven-leancode: fixed in both files.

EXAMPLES.md
aydinguven-leancode: Hi, while the examples changed drastically, the point having the fieldController as a part of builder params makes sense. I have prepared a PoC on how this could work, here is the PR about that: https://github.com/leancodepl/leancode_forms/pull/43

the PR is breaking things and having a conflict but if you could share your opinion on the core idea of using <C> instead of <T, E> that would be already great.

example/lib/screens/complex_form.dart
```diff
@@ -26,36 +28,51 @@ class ComplexForm extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final controller =
+        context.select<ComplexFormController, ComplexFormController>((c) => c);
```
PiotrRogulski: Use `watch` when getting the entire value

lib/src/field/advanced_text_field_controller.dart
```diff
@@ -0,0 +1,61 @@
...
+    if (textController.text != value.value) {
+      textController.text = value.value;
+    }
+  }
+
+  /// Requests focus for the field via [focusNode].
+  void focus() => focusNode.requestFocus();
+
+  /// Clears the value of the field, resetting it to its initial value.
+  void clear() => reset();
```
PiotrRogulski: Do we need another method to reset the field? This might cause confusion for users ("Should I call `clear` or `reset`?")

lib/src/form/advanced_form_controller.dart
```diff
@@ -0,0 +1,421 @@
...
+class AdvancedFormController
+    with ChangeNotifier
+    implements ValueListenable<AdvancedFormState> {
+  /// Creates a new [AdvancedFormController].
+  AdvancedFormController({
+    this.debugName = '',
+    this.validateAll = false,
+  }) : _value = const AdvancedFormState();
+
+  AdvancedFormState _value;
```
PiotrRogulski:
```suggestion
  });

  AdvancedFormState _value = const AdvancedFormState();
```

lib/src/form/advanced_form_controller.dart
```diff
@@ -0,0 +1,421 @@
...
+  }
+
+  /// Flag about whether controller has been disposed. Once true it stays true.
+  /// Disposed controllers are not to be reused
+  bool _isDisposed = false;
+
+  /// Getter for the isDisposed flag
+  bool get isDisposed => _isDisposed;
+
+  /// A debug label for this form. Not significant to the form.
```
PiotrRogulski: I'd avoid writing "not significant" in the docs – users will most likely not use it at all or enter *whatever* as the value. Instead, describe its purpose (I don't see any usages in this PR, though 🤔)

lib/src/form/advanced_form_controller.dart
```diff
@@ -0,0 +1,421 @@
...
+
+    _runChildCleanups();
+
+    _setState(AdvancedFormState(
+      wasModified: value.wasModified,
+      fields: value.fields,
+      subforms: {...value.subforms, form},
+      validationEnabled: value.validationEnabled,
+      validating: value.validating,
+    ),);
```
PiotrRogulski: Reformat

lib/src/field/advanced_text_field_controller.dart
aydinguven-leancode: Removed the clear() method since it just calls the reset()

I see a value of adding it later, as to clear the field while reset brings it back to the initial value.

lib/src/form/advanced_form_controller.dart
aydinguven-leancode: Removed the bit of the comment with "Not significant" and added more info about its usage.

lib/src/form/advanced_form_controller.dart
aydinguven-leancode: Done

lib/src/form/advanced_form_controller.dart
aydinguven-leancode: Done

example/lib/screens/complex_form.dart
aydinguven-leancode: That's right. Changed.

lib/src/field/advanced_field_controller.dart
```diff
@@ -0,0 +1,388 @@
...
+  })  : _value = AdvancedFieldState<T, E>(value: initialValue),
+        _initialValue = initialValue,
+        _validator = validator ?? ((_) => null),
+        _asyncValidator = asyncValidator,
+        _asyncValidationDebounce = asyncValidationDebounce;
+
+  /// Optional name for the field. Useful for debugging, logging, and
+  /// serialization. Not used for identity — fields are still identified by
+  /// reference.
+  final String? name;
```
PiotrRogulski: I think that maybe an example with these names would be useful; I'm still not sure how to use this field

mateusz-pietras: What kind of example did you have in mind? Smth like this?
```suggestion
  /// Optional name for the field. Useful for debugging, logging, and
  /// serialization. Not used for identity — fields are still identified by
  /// reference. Examples: `username`, `address.city`, `privacyPolicyCheckbox`.
  final String? name;
```

example/lib/screens/complex_form.dart
```diff
@@ -14,8 +16,8 @@ class ComplexFormScreen extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return BlocProvider<ComplexFormCubit>(
-      create: (context) => ComplexFormCubit(),
+    return ChangeNotifierProvider<ComplexFormController>(
+      create: (_) => ComplexFormController(),
```
PiotrRogulski: let's not ignore context params 😔 

replace every `(_)` with `(context)` 🙏

example/lib/screens/complex_form.dart
```diff
@@ -122,35 +139,42 @@ class DogSubform extends StatelessWidget {
...
 
-  final dogSubform = DogSubformCubit();
+  final dogSubform = DogSubformController();
+  final humanSubform = HumanSubformController();
+
+  Timer? _typeDebounce;
+  SubformType? _lastSeenType;
 
-  final humanSubform = HumanSubformCubit();
+  void _onTypeListenerFired() {
```
PiotrRogulski: `on...Fired` -> `on...Changed`

example/lib/screens/quiz_form.dart
```diff
@@ -13,44 +13,58 @@ class QuizFormScreen extends StatelessWidget {
...
-class QuizForm extends HookWidget {
+class QuizForm extends StatelessWidget {
   const QuizForm({super.key});
 
   @override
   Widget build(BuildContext context) {
-    final formStatus = context.select<QuizCubit, ValidationStatus>(
-      (cubit) => cubit.state.validationStatus,
+    final controller = context.watch<QuizController>();
+    final formStatus = context.select<QuizController, ValidationStatus>(
```
PiotrRogulski: Read from `controller` directly

lib/src/field/advanced_multi_select_field_controller.dart
```diff
@@ -0,0 +1,35 @@
...
+
+  /// Adds the given [value].
+  void addValue(V value) {
+    setValue(Set<V>.from(this.value.value)..add(value));
+  }
+
+  /// Removes the given [value].
+  void removeValue(V value) {
+    setValue(Set<V>.from(this.value.value)..remove(value));
+  }
```
PiotrRogulski, line 40:
```suggestion
  void addValue(V value) {
    setValue({...this.value.value}..add(value));
  }

  /// Removes the given [value].
  void removeValue(V value) {
    setValue({...this.value.value}..remove(value));
  }
```

Alternatively, `.toSet()` or `Set.of()`

lib/src/field/advanced_field_controller.dart
```diff
@@ -0,0 +1,498 @@
...
+
+  /// Validates the field's value. Returning null means valid.
+  final AsyncValidator<T, E> validator;
+
+  /// How long to wait after a value change before running [validator]. Each
+  /// new value restarts the wait, so a burst of changes runs [validator] once
+  /// rather than once per change.
+  final Duration debounce;
+
+  /// Called when [validator] throws. If ommitted, error is reported through
```
PiotrRogulski:
```suggestion
  /// Called when [validator] throws. If omitted, error is reported through   
```

lib/src/field/advanced_field_controller.dart
PiotrRogulski: I was more thinking of the ***usage*** of these names.

> Useful for debugging, logging, and serialization

Would be nice to have a code example for this

lib/src/field/advanced_field_controller.dart
```diff
@@ -0,0 +1,498 @@
...
+        readOnly,
+        status,
+      );
+}
+
+class _Unset {
+  const _Unset();
+}
+
+const _unset = _Unset();
```
PiotrRogulski:
```suggestion
const _unset = Object();
```
This should be enough

lib/src/field/advanced_field_controller_extensions.dart
PiotrRogulski, line 1: Are these really needed and do they bring value? These effectively make our forms have three APIs for listening

MIGRATION.md
```diff
@@ -0,0 +1,491 @@
...
+| `form.onValuesChangedStream` (`Stream<void>`) | `form.onValuesChanged` (`Listenable`) |
+| `form.onStatusChangedStream` (`Stream<FieldStatus>`) | `form.onStatusChanged` (`Listenable`) |
+| `Future<void> close()` | `void dispose()` |
+
+And in your `pubspec.yaml`:
+
+- You can **drop `flutter_bloc` and `rxdart`** — unless something else in your app still needs them, of course.
+- If your tests or DI used **`bloc_test`, `bloc_presentation`, or `flutter_hooks`** just for forms, those can go too. **Add `provider`** if you want a drop-in replacement for `BlocProvider`/`context.read` (see [section 8](#8-migrating-exampletest-code-that-used-flutter_bloc-directly)).
+
+Fewer dependencies is the point, not a side effect: a forms library shouldn't be the reason your lockfile has a stream toolkit in it.
```
PiotrRogulski: Doesn't seem necessary for a migration doc 😛 I'm guessing folks reading this expect a step-by-step, not a promo 🙈

(I'd just let an LLM revise this entire doc into a more technical writing)

MIGRATION.md
```diff
@@ -0,0 +1,491 @@
...
+  field.dispose();
+  super.dispose();
+}
+```
+
+> **Why was `close()` async in the first place?** Because it closed a broadcast `StreamController` under the hood. `ChangeNotifier.dispose()` just clears a listener list — nothing to await, in the library or in your code.
+
+### One thing to actually watch out for
+
+`dispose()` is not idempotent: call it twice on a notifier and debug mode throws `'A ValueNotifier was used after being disposed'`. The old `Cubit.close()` silently tolerated a re-close — so if your 0.1.x code disposed a field by hand *and* let `FormGroupCubit.close()` dispose it again, that worked purely by accident, and 0.2.0 will call you out on it.
```
PiotrRogulski:
```suggestion
`dispose()` is not idempotent: call it twice on a notifier in debug mode throws `'A ValueNotifier was used after being disposed'`. The old `Cubit.close()` silently tolerated a re-close — so if your 0.1.x code disposed a field by hand *and* let `FormGroupCubit.close()` dispose it again, that worked purely by accident, and 0.2.0 will call you out on it.
```

lib/src/field/advanced_field_controller_extensions.dart
mateusz-pietras: Right, I'd leave `stream` extension to make the migration from `bloc` easier. We can mark it as `@Deprecated` wdyt?

MIGRATION.md
mateusz-pietras: Good point

lib/src/field/advanced_field_controller_extensions.dart
PiotrRogulski: Yeah, deprecation sounds ok 👍

test/src/field/field_controller_extensions_test.dart
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  group('stream', () {
+    test('emits states after subscription', () async {
+      final emissions = <AdvancedFieldState<_Value, _Error>>[];
+      final subscription = field.stream.listen(emissions.add);
+      // Broadcast stream attaches its listener on onListen, which is
+      // synchronous here, but events are delivered asynchronously.
+      field.setValue((unit: 'kg', amount: 2));
+      await Future<void>.delayed(Duration.zero);
```
PiotrRogulski: Use `pumpEventQueue` instead of zero-time delays

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+* **Breaking:** Lifecycle method renamed from `close()` to `dispose()` on both controllers.
+* **Breaking:** `FormController` exposes `onValuesChanged` and `onStatusChanged` as `Listenable`s (previously `Stream`s named `onValuesChangedStream` / `onStatusChangedStream`).
+* `TextFieldController` now owns a `TextEditingController` (`field.textController`) kept in two-way sync with the field value. Widgets bind to it directly; programmatic changes (`setValue`, `reset`, `clear`) propagate to the text controller, and user input propagates back. Removes the dual-write bug class around resetting fields, set-to-initial flows, etc.
+* `TextFieldController` now owns a `FocusNode` (`field.focusNode`) and exposes a `focus()` shortcut. Enables scroll-to-first-invalid, sequential focus navigation, and programmatic focus from validators without subclassing.
+* `FieldController` gained an optional `String? name` parameter for debugging, logging, and serialization.
+* `Disposable` mixin removed (lifecycle is handled by `ChangeNotifier.dispose`).
+* Internal `distinctWithFirst` stream extension and `CancelableFuture` are no longer exported.
```
PiotrRogulski: I don't think they were even exported 🤔 (at least they don't show up in the latest docs)

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+## 0.2.0
+
+> Upgrading from 0.1.x? See [MIGRATION.md](./MIGRATION.md) for a step-by-step guide.
+
+* **Breaking:** We've rebuilt the library on `ValueNotifier` / `ChangeNotifier`. That means `flutter_bloc` and `rxdart` no longer are a dependency.
```
PiotrRogulski:
```suggestion
* **Breaking:** We've rebuilt the library on `ValueNotifier` / `ChangeNotifier`. That means `flutter_bloc` and `rxdart` are no longer dependencies.
```

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+* **Breaking:** Lifecycle method renamed from `close()` to `dispose()` on both controllers.
+* **Breaking:** `FormController` exposes `onValuesChanged` and `onStatusChanged` as `Listenable`s (previously `Stream`s named `onValuesChangedStream` / `onStatusChangedStream`).
+* `TextFieldController` now owns a `TextEditingController` (`field.textController`) kept in two-way sync with the field value. Widgets bind to it directly; programmatic changes (`setValue`, `reset`, `clear`) propagate to the text controller, and user input propagates back. Removes the dual-write bug class around resetting fields, set-to-initial flows, etc.
+* `TextFieldController` now owns a `FocusNode` (`field.focusNode`) and exposes a `focus()` shortcut. Enables scroll-to-first-invalid, sequential focus navigation, and programmatic focus from validators without subclassing.
+* `FieldController` gained an optional `String? name` parameter for debugging, logging, and serialization.
+* `Disposable` mixin removed (lifecycle is handled by `ChangeNotifier.dispose`).
```
PiotrRogulski, line 23: Similarly, this was an implementation detail, so no need to mention it 🤔

README.md
PiotrRogulski, line 1: I'm not exactly sure how else to explain it, but this readme is now very ***wordy*** — I think we should make it more concise

CHANGELOG.md
mateusz-pietras: It wasn't only internal though — `FormGroupCubit` had `with Disposable`, and `addDisposable` was `@protected`, so any form subclass could call it. 

Our own example did: `complex_form.dart` used `addDisposable(type.stream.debounceTime(...).listen(...).cancel) `.

I'd leave it but reword the changelog mention

README.md
mateusz-pietras: Yup, sure thing, will revise

CHANGELOG.md
PiotrRogulski: Ah, I see 👍

test/src/form/form_controller_test.dart
```diff
@@ -0,0 +1,810 @@
...
+        ]);
+        form.dispose();
+      });
+
+      test('does not change if field was unregistered', () async {
+        form
+          ..registerFields([field1, field2])
+          ..registerFields([]);
+
+        await Future<void>.delayed(Duration.zero);
```
PiotrRogulski: There are many more zero-time delays – replace all globally with pumpEventQueue

PiotrRogulski (approved):

example/lib/screens/password_form.dart
```diff
@@ -15,8 +16,8 @@ class PasswordFormScreen extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return BlocProvider<PasswordFormCubit>(
-      create: (context) => PasswordFormCubit(),
+    return ChangeNotifierProvider<PasswordFormController>(
```
KamilSztandur, line 19: I wonder if we could have `FormProvider` widget, which would do exactly the same, but also have shorter and more form-related name.

lib/src/field/advanced_single_select_field_controller.dart
```diff
@@ -0,0 +1,20 @@
...
+    super.validator,
+    required this.options,
+    super.name,
+  });
+
+  /// List of options to select from.
+  final List<V> options;
+
+  /// Sets the value of the field to the [option].
+  void select(V? option) => setValue(option);
```
KamilSztandur: I think we need additional check whether `option` is part of `options` list.

lib/src/field/advanced_multi_select_field_controller.dart
```diff
@@ -0,0 +1,35 @@
...
+    if (fieldValue.contains(value)) {
+      removeValue(value);
+    } else {
+      addValue(value);
+    }
+  }
+
+  /// Adds the given [value].
+  void addValue(V value) {
+    setValue({...fieldValue}..add(value));
```
KamilSztandur, line 34: I think we need additional check whether `value` is on `options` list

KamilSztandur (approved):

KamilSztandur:
I think that we'll need to rework the `/example/` section. There are a lot of antypatterns like widget-returning methods, which makes the library looks like it doesn't care about code quality 😅 most of incoming devs won't examine the implementation itself

Most of the widgets here also seem a bit overcomplicated with a lot of magic numbers and fancy designs. Let's keep widgets in `/example/` as simple as possible. We want to make this library look time-saving and easy-to-use. Currently, there's too much noise related to these designs, not the advanced_forms usage itself

But let's do that as part of README/DOC update task. Great work here <3

example/lib/screens/password_form.dart
mateusz-pietras: Moved to a separate ticket

PR #43, opened 2026-06-23
PoC on type definition change

aydinguven-leancode:


mateusz-pietras:
POC rejected - makes type definition too complex

Issue #28, opened 2025-05-30
FormValidationStatusBuilder


piotruela:
I'd like to have a builder wrapper, that triggers the rebuild whenever validation status of the form changes. 
In my use case I want to autovalidate the form on each change and I want to enable/disable the submit button based on the validation status of the form.
For now I created a such wrapper:
```dart
class FormValidationStatusBuilder extends StatelessWidget {
  const FormValidationStatusBuilder({
    super.key,
    required this.formCubit,
    required this.builder,
  });

  final FormGroupCubit formCubit;
  final Widget Function(BuildContext context, bool isValid) builder;

  @override
  Widget build(BuildContext context) {
    return StreamBuilder(
      stream: formCubit.onStatusChangedStream,
      builder: (context, _) {
        return builder(
          context,
          formCubit.state.allFields.every((field) => field.state.isValid),
        );
      },
    );
  }
}
```

I think this case should be handled by package ootb

KamilSztandur:
We decided to not introduce this widget, because it's a bit misleading. It sounds like submit button should be automatically toggled depending on the validation, but UI won't refresh if form isn't in `autovalidate` mode. 

We could make every form automatically validated once this widget is in the tree, but we don't want to control business logic via UI.

PR #48, opened 2026-06-24
Added AdvancedValidationStatusBuilder

aydinguven-leancode:
Closes [[LMG-398](https://leancode.atlassian.net/browse/LMG-398)]
Closes issue #28 

Added AdvancedValidationStatusBuilder + tests + new example screen

[LMG-398]: https://leancode.atlassian.net/browse/LMG-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -7,6 +7,7 @@ import 'package:leancode_forms_example/screens/password_form.dart';
 import 'package:leancode_forms_example/screens/quiz_form.dart';
 import 'package:leancode_forms_example/screens/scroll_form.dart';
 import 'package:leancode_forms_example/screens/simple_form.dart';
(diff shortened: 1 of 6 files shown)


mateusz-pietras:
Rejecting because AdvancedValidationStatusBuilder gates Submit on AdvancedFormState.isValid, which reads each field's stored status — and valid is the default, not a verdict, so the button is enabled over an untouched empty required form, never rebuilds while the user types (with autovalidate off no status transition occurs, so onStatusChanged never fires), and is re-enabled by resetAll(). The gate needs to be derived from the current values (run the sync validators, honour validationEnabled, and listen to onValuesChanged as well) rather than from a status that nothing has written yet.

PR #50, opened 2026-06-24
[LMG-401] Debouncer audit

aydinguven-leancode:
Covered the case where debouncer leaks on disposed fields

### DIFF
--- a/lib/src/field/advanced_field_controller.dart
+++ b/lib/src/field/advanced_field_controller.dart
@@ -140,10 +140,9 @@ class AdvancedFieldController<T, E extends Object>
     notifyListeners();
   }
 
-  /// Whether this controller has been disposed. Once true it stays true.
   bool _isDisposed = false;
 
-  /// Whether this controller has been disposed.
+  /// Whether this controller has been disposed. Once true it stays true.
   bool get isDisposed => _isDisposed;
 
   /// Test-only seeder for states the public API can't construct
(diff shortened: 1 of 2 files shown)


lib/src/field/advanced_field_controller.dart
```diff
@@ -140,6 +140,12 @@ class AdvancedFieldController<T, E extends Object>
     notifyListeners();
   }
 
+  /// Whether this controller has been disposed. Once true it stays true.
+  bool _isDisposed = false;
+
+  /// Whether this controller has been disposed.
```
PiotrRogulski: I'd either remove the private comment, or at least unify/reuse

test/src/field/field_controller_test.dart
```diff
@@ -254,6 +291,139 @@ void main() {
...
+        _State(value: 10, status: FieldStatus.pending),
+        _State(value: 10, status: FieldStatus.validating),
+      ]);
+    });
+
+    // `testWidgets` on purpose: the binding fails a test that leaves a pending
+    // `Timer` behind, which is exactly the leak being guarded against.
+    testWidgets(
+        'disposing from a listener during the pending emission does not '
+        'schedule a debounce timer', (tester) async {
```
PiotrRogulski: Q: do we want to support this in any case? Disposing from a listener is "illegal", so I'm not sure if we need to make sure the package behaves correctly

mateusz-pietras: woops, true, we shouldn't include that, good catch

KamilSztandur (approved):

PR #51, opened 2026-08-06
feat: impl custom form provider proposal

mateusz-pietras:
create a simple Provider implementation with limited set of features to remove dependency on provider package; update docs and examples to use custom provider

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * **Deprecated:** `AdvancedFieldController.stream` replaces the `stream` that `FieldCubit` inherited from `Cubit`, so stream-based code keeps compiling through the migration. It ships deprecated and will be removed in 0.3.0 — use `addListener`, `subscribeToFields`, or the builder widgets instead.
 * **Breaking:** Removed `clear()` from the text, single-select, and multi-select controllers — it only called `reset()`. Call `reset()` instead.
 * **Breaking:** Dropped the `equatable` dependency. `AdvancedFieldState` and `AdvancedFormState` now implement `==` / `hashCode` by hand, comparing members with `==` rather than `EquatableMixin`'s deep collection equality. Identical for scalar and record values; for `List` / `Set` / `Map` values or error types, two equal-content-but-distinct instances now compare unequal, so a set-to-equal-value notifies where it previously deduplicated.
+* Added `AdvancedFormScope`, a small `StatefulWidget` for creating, exposing, and disposing an `AdvancedFormController` — no dependency on the `provider` package required. `create` runs lazily on first `AdvancedFormScope.watch` / `.read`, and the controller is disposed automatically on unmount. See [README.md](./README.md) and [MIGRATION.md](./MIGRATION.md#8-dropping-flutter_bloc-and-friends). Already using `provider`? `ChangeNotifierProvider` still works exactly the same way — use whichever you prefer.
 
 ## 0.1.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,9 @@
 * **Deprecated:** `AdvancedFieldController.stream` replaces the `stream` that `FieldCubit` inherited from `Cubit`, so stream-based code keeps compiling through the migration. It ships deprecated and will be removed in 0.3.0 — use `addListener`, `subscribeToFields`, or the builder widgets instead.
 * **Breaking:** Removed `clear()` from the text, single-select, and multi-select controllers — it only called `reset()`. Call `reset()` instead.
 * **Breaking:** Dropped the `equatable` dependency. `AdvancedFieldState` and `AdvancedFormState` now implement `==` / `hashCode` by hand, comparing members with `==` rather than `EquatableMixin`'s deep collection equality. Identical for scalar and record values; for `List` / `Set` / `Map` values or error types, two equal-content-but-distinct instances now compare unequal, so a set-to-equal-value notifies where it previously deduplicated.
-* Added `AdvancedFormScope`, a small `StatefulWidget` for creating, exposing, and disposing an `AdvancedFormController` — no dependency on the `provider` package required. `create` runs lazily on first `AdvancedFormScope.watch` / `.read`, and the controller is disposed automatically on unmount. See [README.md](./README.md) and [MIGRATION.md](./MIGRATION.md#8-dropping-flutter_bloc-and-friends). Already using `provider`? `ChangeNotifierProvider` still works exactly the same way — use whichever you prefer.
+* Added `AdvancedFormScope`, which creates, exposes, and disposes an `AdvancedFormController` without the `provider` package. `create` runs lazily on first `.watch` / `.read`, and the controller is disposed on unmount. `ChangeNotifierProvider` still works if you prefer it — see [MIGRATION.md](./MIGRATION.md#8-dropping-flutter_bloc-and-friends).
+* Added `context.watchForm<T>()` and `context.readForm<T>()` as shorthands for `AdvancedFormScope.watch` / `.read`.
+* Added `AdvancedFormScopeNotFoundException`, thrown when a lookup finds no matching scope.
 
 ## 0.1.2
 
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -60,11 +60,11 @@ AdvancedFormScope<SimpleFormController>(
   child: const SimpleForm(),
 )
 
-// Inside the SimpleForm widget — AdvancedFormScope.read is the right call
-// here, because we want a stable reference for the onPressed callback (no
-// subscription to rebuilds).
+// Inside the SimpleForm widget — readForm is the right call here, because we
+// want a stable reference for the onPressed callback (no subscription to
+// rebuilds).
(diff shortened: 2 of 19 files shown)


lib/src/form/advanced_form_scope.dart
```diff
@@ -0,0 +1,196 @@
...
+    }
+    return inherited.state._ensureController();
+  }
+
+  /// Looks the controller up without subscribing to rebuilds — `context`
+  /// will not rebuild when the controller notifies its listeners.
+  ///
+  /// Throws a [StateError] if no matching scope is found.
+  static T read<T extends AdvancedFormController>(BuildContext context) {
+    final element = context
```
PiotrRogulski: Why not `getInheritedWidgetOfExactType`? You wouldn't need to go through the element and then back to the widget

lib/src/form/advanced_form_scope.dart
```diff
@@ -0,0 +1,196 @@
...
+///
+/// // Elsewhere, below the scope:
+/// final controller = AdvancedFormScope.watch<MyFormController>(context);
+/// ```
+class AdvancedFormScope<T extends AdvancedFormController>
+    extends StatefulWidget {
+  /// Creates a scope that builds its controller lazily, on first lookup, via
+  /// [create]. The controller is disposed when this widget unmounts.
+  const AdvancedFormScope({
+    required this.create,
```
PiotrRogulski: `create` here should be nonnullable

lib/src/form/advanced_form_scope.dart
```diff
@@ -0,0 +1,196 @@
...
+      version: _version,
+      child: widget.child,
+    );
+  }
+}
+
+/// Carries the version counter [AdvancedFormScope.watch] subscribes to, and
+/// a reference to the [State] that owns the controller, which
+/// [AdvancedFormScope.read] reads directly without subscribing.
+class _AdvancedFormScopeMarker<T extends AdvancedFormController>
```
PiotrRogulski: Why the `Marker` suffix? I feel like this can just be called `_AdvancedFormScope` (see https://github.com/search?q=repo%3Aflutter%2Fflutter+%2Fextends+InheritedWidget%2F&type=code)

lib/src/form/advanced_form_scope.dart
```diff
@@ -0,0 +1,196 @@
...
+    final element = context
+        .getElementForInheritedWidgetOfExactType<_AdvancedFormScopeMarker<T>>();
+    final widget = element?.widget as _AdvancedFormScopeMarker<T>?;
+    if (widget == null) {
+      throw _notFoundError<T>();
+    }
+    return widget.state._ensureController();
+  }
+
+  static StateError _notFoundError<T>() => StateError(
```
PiotrRogulski: I don't think this should be an error, but an exception. Also, I'd define a dedicated class for this (like in https://github.com/rrousselGit/provider/blob/6cfb56aec9f405dae0030d78f43bf4817265c5eb/packages/provider/lib/src/provider.dart#L479)

example/lib/screens/scroll_form.dart
```diff
@@ -34,11 +33,12 @@ class _ScrollFormState extends State<ScrollForm> {
...
   void didChangeDependencies() {
     super.didChangeDependencies();
-    _eventSubscription ??= context.read<ScrollFormController>().events.listen(
+    _eventSubscription ??=
+        AdvancedFormScope.read<ScrollFormController>(context).events.listen(
       (event) {
         final context = this.context;
         if (event is SubmitFailedWithErrors && context.mounted) {
-          _scrollToFirstError(context.read<ScrollFormController>());
+          _scrollToFirstError(AdvancedFormScope.read<ScrollFormController>(context));
```
PiotrRogulski: It's a bummer that it's much longer now; maybe we can also ship an extension on context to simplify this (maybe `readForm` & `watchForm`)?

example/lib/screens/complex_form.dart
```diff
@@ -54,12 +53,11 @@ class ComplexForm extends StatelessWidget {
               labelText: 'Subform Type',
               hintText: 'Select subform type',
             ),
-            Builder(
+            Builder(
```
PiotrRogulski: reformat files 🙏 (as part of the migration, we could also set up CI checks 😛)

test/src/form/advanced_form_scope_test.dart
```diff
@@ -0,0 +1,321 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:leancode_forms/leancode_forms.dart';
+
+/// A controller that records whether/how many times it was created and
```
PiotrRogulski: > that records whether/how many times it was created

not really, it only counts disposes 😛 (you do count `createCalls`, but it's in a different place)

test/src/form/advanced_form_scope_test.dart
```diff
@@ -0,0 +1,321 @@
...
+    });
+
+    testWidgets('throws a StateError with a helpful message when not found',
+        (tester) async {
+      await tester.pumpWidget(
+        MaterialApp(
+          home: _Reader(
+            builder: (context) {
+              expect(
+                () => AdvancedFormScope.watch<_RecordingController>(context),
```
PiotrRogulski: I'm not too fond of placing an expect "deep" in a widget tree 🤔 Maybe we can put the method call normally, let the exception happen, and assert it in the test body? (there's `tester.takeException`)

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
...
 * **Breaking:** Removed `clear()` from the text, single-select, and multi-select controllers — it only called `reset()`. Call `reset()` instead.
 * **Breaking:** Dropped the `equatable` dependency. `AdvancedFieldState` and `AdvancedFormState` now implement `==` / `hashCode` by hand, comparing members with `==` rather than `EquatableMixin`'s deep collection equality. Identical for scalar and record values; for `List` / `Set` / `Map` values or error types, two equal-content-but-distinct instances now compare unequal, so a set-to-equal-value notifies where it previously deduplicated.
+* Added `AdvancedFormScope`, a small `StatefulWidget` for creating, exposing, and disposing an `AdvancedFormController` — no dependency on the `provider` package required. `create` runs lazily on first `AdvancedFormScope.watch` / `.read`, and the controller is disposed automatically on unmount. See [README.md](./README.md) and [MIGRATION.md](./MIGRATION.md#8-dropping-flutter_bloc-and-friends). Already using `provider`? `ChangeNotifierProvider` still works exactly the same way — use whichever you prefer.
```
PiotrRogulski: A bit too wordy for my taste 🙈 (for example, we prolly don't need to say that it's stateful)

EXAMPLES.md
```diff
@@ -93,10 +92,10 @@ class SignupController extends AdvancedFormController {
...
 }
 
-// Let's select field reference
-final email = context.select<SignupController, AdvancedTextFieldController<ValidationError>>(
-  (c) => c.email,
-);
+// Grab the field reference — no select equivalent, but a field reference is
+// stable across rebuilds, so a plain read is enough; AdvancedFieldBuilder
+// below is what actually subscribes to the field's own state.
+final email = AdvancedFormScope.read<SignupController>(context).email;
```
PiotrRogulski: This is an unfortunate example, because it now "violates" a rule from Provider — about using `read` directly in build. This should use `watch` (or could we implement `select` as well?)

example/lib/screens/scroll_form.dart
mateusz-pietras: That's worth adding, thanks

lib/src/form/advanced_form_scope.dart
```diff
@@ -0,0 +1,220 @@
...
+  /// Looks the controller up without subscribing to rebuilds — `context`
+  /// will not rebuild when the controller notifies its listeners.
+  ///
+  /// Throws an [AdvancedFormScopeNotFoundException] if no matching scope is
+  /// found.
+  static T read<T extends AdvancedFormController>(BuildContext context) {
+    final scope =
+        context.getInheritedWidgetOfExactType<_AdvancedFormScope<T>>();
+    if (scope == null) {
+      throw AdvancedFormScopeNotFoundException(T);
```
PiotrRogulski, line 82: Why not pass `T` as a generic param?


```suggestion
      throw AdvancedFormScopeNotFoundException<T>();
```

test/src/form/advanced_form_scope_test.dart
```diff
@@ -0,0 +1,363 @@
...
+
+      controllerB.setValidationEnabled(false);
+      await tester.pump();
+      expect(buildCalls, 3);
+
+      expect(controllerA.isDisposed, isFalse);
+      expect(controllerB.isDisposed, isFalse);
+    });
+
+    testWidgets('nested scopes of the same type resolve to the innermost',
```
PiotrRogulski, line 243: tbh, I'm not sure how this tests ensures that the read controller is indeed from the innermost scope (I'd expect maybe inspecting both elements and comparing their states? 🤔)

mateusz-pietras:
Closed - excluded from the scope, leaves `provider` in docs as default suggested DI

PR #53, opened 2026-08-11
feat(validation): refine validation pipeline; fix validation issues

mateusz-pietras:
- add support for async validation when not using `autovalidate`, triggered by validate
- rewrite validation pipeline

Reviewer note:
- don't put much attention to the .md files - they'll get rewritten as the last step of this rework, before publishing


example/lib/screens/simple_form.dart
```diff
@@ -96,19 +129,28 @@ class SimpleFormController extends AdvancedFormController {
...
+          ValidationError.emailCheckUnavailable,
     ),
   );
 
-  Future<ValidationError?> _onEmailChanged(String value) async {
+  Future<ValidationError?> _checkEmail(String value) async {
     final takenEmail = ['john@email.com', 'jack@email.com'];
     await Future<void>.delayed(const Duration(milliseconds: 700));
+    if (value == 'boom@email.com') {
+      throw StateError('the email service is down');
```
PiotrRogulski: why `StateError`? I'd use some exception

lib/src/field/advanced_field_state.dart
```diff
@@ -0,0 +1,144 @@
+part of 'advanced_field_controller.dart';
+
+/// Translates an error to a string.
+typedef ErrorTranslator<E extends Object> = String Function(E);
+
+/// The status of a [AdvancedFieldController].
+enum FieldStatus {
+  /// No error is recorded on the field.
+  ///
+  /// Not the same as *checked and passed*: a field nobody has validated yet is
```
PiotrRogulski, line 10: > Not the same as *checked and passed*

But there isn't a separate status for that, though 🤔

lib/src/field/advanced_field_state.dart
```diff
@@ -0,0 +1,144 @@
...
+
+  /// The current error: [validationError] falling back to [asyncError].
+  ///
+  /// [validationError] wins because a rule about the value outranks a verdict
+  /// that may predate the app state the rule reads.
+  E? get error => validationError ?? asyncError;
+
+  /// Whether no error is recorded. See [FieldStatus.valid]: this is not the
+  /// same as *checked and passed*.
+  bool get isValid => status == FieldStatus.valid;
```
PiotrRogulski: Ditto — there's also no getter to check for `checked and passed`. I'd assume `isValid` means exactly this

lib/src/field/advanced_field_state.dart
```diff
@@ -0,0 +1,144 @@
...
+      'validationError: $validationError, '
+      'asyncError: $asyncError, '
+      'status: ${status.name}, '
+      'autovalidate: $autovalidate, '
+      'readOnly: $readOnly)';
+}
+
+enum _Unset { unset }
+
+const _unset = _Unset.unset;
```
PiotrRogulski:
```suggestion
const _unset = Object();
```

lib/src/field/async_validation.dart
```diff
@@ -0,0 +1,135 @@
...
+  /// cannot turn a failed round into a passing one.
+  final AsyncValidationFailureHandler? onFailure;
+
+  /// Turns a failure into an error code, so a failed round can show something.
+  /// Without it a failed field carries no code. A code produced here lands in
+  /// [AdvancedFieldState.asyncError].
+  final AsyncValidationFailureMapper<E>? failureToError;
+}
+
+extension _Failures<T, E extends Object> on AsyncValidation<T, E> {
```
PiotrRogulski: This can be regular members in the class

lib/src/field/async_validation.dart
```diff
@@ -0,0 +1,135 @@
...
+  E? mapFailure(AsyncValidationFailure failure, String? field) {
+    if (failureToError case final map?) {
+      try {
+        return map(failure.error, failure.stackTrace);
+      } catch (error, stack) {
+        _report(field, 'mapping the async validation failure', error, stack);
+      }
+    }
+    return null;
```
PiotrRogulski:
```suggestion
    try {
      return failureToError?.call(failure.error, failure.stackTrace);
    } catch (error, stack) {
      _report(field, 'mapping the async validation failure', error, stack);
    }
```

lib/src/field/advanced_text_field_controller.dart
```diff
@@ -18,37 +21,61 @@ class AdvancedTextFieldController<E extends Object>
...
+    if (textController.text == text) {
+      return;
     }
+
+    final selection = textController.selection;
+    final newSelection = selection.isValid &&
+            selection.start <= text.length &&
+            selection.end <= text.length
+        ? selection
+        : TextSelection.collapsed(offset: text.length);
```
PiotrRogulski: Question — does such a case behave correctly:

```
              vvvvvvvv selection
|lorem ipsum, dolor sit amet|
```
When the text changes to `lorem ip, dolor sit amet`, I'd expect the selection to remain `dolor si`. But this snippet, from what I understand, will preserve the offset of the selection, which would become `or sit a` 🤔

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
+  /// disposed controllers are not to be reused.
   bool get isDisposed => _isDisposed;
 
   /// Fires when any leaf field's value changes (recursively through subforms),
   /// or when fields are registered.
   Listenable get onValuesChanged => _onValuesChanged;
-  final ChangeNotifier _onValuesChanged = ChangeNotifier();
 
   /// Fires when any leaf field's status changes (recursively through subforms).
   Listenable get onStatusChanged => _onStatusChanged;
```
PiotrRogulski: Hmm, on one hand it's nice that instance fields are now together, but I also liked that the private storage is right next to the public getter 🤔

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
+  /// [enableAutovalidate] is the one that applies. Returns `true` when
   /// `state.validationEnabled` is false.
-  bool validate({bool enableAutovalidate = true}) {
+  Future<bool> validate({bool enableAutovalidate = true}) {
+    // Checked before anything is broadcast: a coalesced call must not re-run
+    // `setAutovalidate`, and a disabled form must not claim the coalescing slot
+    // — a later, enabled call would then return this one's trivial `true`.
+    if (_validateCall.inFlight case final inFlight?) {
+      return inFlight;
+    }
```
PiotrRogulski, line 120: Why not put the entire body of `validate` inside a `_validateCall.run`? This `if` duplicated the logic already inside `CoalescingCall`

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
+    if (isDisposed) {
+      return Future.value(false);
+    }
+
     if (enableAutovalidate) {
       setAutovalidate(true);
     }
     if (!value.validationEnabled) {
-      return true;
+      return Future.value(true);
```
PiotrRogulski: Consider using `Future.syncValue` when possible

lib/src/form/advanced_form_controller.dart
```diff
@@ -349,18 +346,33 @@ class AdvancedFormState {
...
+
+  /// Every error in the tree, keyed by field.
+  ///
+  /// Keyed on [AdvancedFieldState.error], so a field invalid from an async
+  /// check appears too. A failed round with no error code does not.
   Map<AdvancedFieldController<dynamic, dynamic>, dynamic>
       get validationErrors => {
             for (final field in allFields)
-              if (field.value.validationError case final error?) field: error,
+              if (field.value.error case final error?) field: error,
```
PiotrRogulski, line 419:
```suggestion
              field: ?field.value.error,
```

lib/src/field/advanced_field_controller.dart
```diff
@@ -1,62 +1,37 @@
...
+});
 
-/// A single form field which can be validated.
-/// Stores the current value, error text, and whether autovalidate is on.
-/// [T] is the held value, [E] is the type of an error. [E] cannot be nullable
-/// to be able to unambiguously detect lack of errors.
+FieldStatus _slotStatus(Object? validationError, Object? asyncError) =>
+    (validationError ?? asyncError) == null
+        ? FieldStatus.valid
+        : FieldStatus.invalid;
```
KamilSztandur: Maybe just

```dart
validationError == null && asyncError == null
? FieldStatus.valid
: FieldStatus.invalid
```

:D

lib/src/field/advanced_field_controller.dart
```diff
@@ -67,471 +42,455 @@ class AdvancedFieldController<T, E extends Object>
...
-    if (onError == null) {
-      _reportAsyncValidationError(
-        error,
-        stackTrace,
-        'running the async validator',
-      );
+    final result = await round.race(asyncValidation);
+    round.disarm();
+
+    // Superseded, cancelled or disposed: a dead round's answer is noise.
```
KamilSztandur: I don't understand this comment

lib/src/field/advanced_field_controller.dart
```diff
@@ -67,471 +42,455 @@ class AdvancedFieldController<T, E extends Object>
...
-    // `validator` runs before `_asyncValidationFuture` is assigned, so
-    // disposal from inside it leaves nothing for `dispose()` to cancel.
-    if (isDisposed) {
+  Future<void> _run(_ValidationRound<T, E> round) async {
+    if (!identical(_currentRound, round)) {
       return;
     }
+    round.disarm();
 
+    final asyncValidation = _asyncValidation!;
```
KamilSztandur: I'd prefer to not use `!`, but hide it behind if condition

lib/src/field/advanced_field_controller.dart
```diff
@@ -67,471 +42,455 @@ class AdvancedFieldController<T, E extends Object>
...
+          StackTrace.current,
+        ),
       );
-}
+    }
 
-const _unset = Object();
+    try {
+      // `Future.sync` sends a synchronous throw down the rejected-future path,
+      // so no round can rest in `validating` with nothing outstanding.
```
KamilSztandur: I don't understand either 😅

lib/src/field/advanced_field_state.dart
```diff
@@ -0,0 +1,144 @@
...
+      );
+
+  @override
+  String toString() => 'AdvancedFieldState('
+      'value: $value, '
+      'validationError: $validationError, '
+      'asyncError: $asyncError, '
+      'status: ${status.name}, '
+      'autovalidate: $autovalidate, '
+      'readOnly: $readOnly)';
```
KamilSztandur: At this point, I wonder if we want to do it manually or just use `equatable` package. It doesn't have many dependencies, so I think we can use it https://github.com/felangel/equatable/blob/master/pubspec.yaml

lib/src/field/advanced_text_field_controller.dart
```diff
@@ -18,37 +21,61 @@ class AdvancedTextFieldController<E extends Object>
...
+
   void _onTextControllerChanged() {
-    if (textController.text != value.value) {
+    if (_reconciling) {
+      return;
+    }
+    if (textController.text != fieldValue) {
       setValue(textController.text);
+      // The field may have refused the write — read-only, or a subclass
+      // transformed the value — so reconcile in the same turn.
```
KamilSztandur: Unnecessary comment

lib/src/utils/coalescing_call.dart
```diff
@@ -0,0 +1,30 @@
+import 'dart:async';
+
+/// Holds the run of an operation that must never be in flight twice at once,
+/// so a double-tapped submit button cannot start two validation passes.
+class CoalescingCall<T> {
```
KamilSztandur: Mark as `@internal`

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
+  /// field whose gate is open. See [validateWithAutovalidate].
   final bool validateAll;
 
-  /// Flag about whether controller has been disposed. Once true it stays true.
-  /// Disposed controllers are not to be reused
+  final ChangeNotifier _onValuesChanged = ChangeNotifier();
+  final ChangeNotifier _onStatusChanged = ChangeNotifier();
+  final Set<AdvancedFieldController<dynamic, dynamic>> _ownedFields = {};
+  final List<VoidCallback> _childCleanups = [];
+  final CoalescingCall<bool> _validateCall = CoalescingCall();
```
KamilSztandur:
```dart
  final _onValuesChanged = ChangeNotifier();
  final _onStatusChanged = ChangeNotifier();
  final _ownedFields = <AdvancedFieldController<dynamic, dynamic>>{};
  final _childCleanups = <VoidCallback>[];
  final _validateCall = CoalescingCall<bool>();
```

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
+
+  AdvancedFormState _value = const AdvancedFormState();
+  List<dynamic> _initialFieldsState = const <dynamic>[];
```
KamilSztandur:
```dart
  var _value = const AdvancedFormState();
  var _initialFieldsState = const <dynamic>[];
```

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
   bool _isDisposed = false;
 
-  /// Getter for the isDisposed flag
+  @override
+  AdvancedFormState get value => _value;
+
+  /// Whether this controller has been disposed. Once true it stays true;
+  /// disposed controllers are not to be reused.
```
KamilSztandur, line 40: I think this comment is unnecessary

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
-      return;
-    }
-    _value = newValue;
-    notifyListeners();
-  }
-
-  /// A label for this form used for logging, or tracing
-  /// across nested subforms. Has no effect on form behavior.
+  /// A label for this form used for logging, or tracing across nested
+  /// subforms. Has no effect on form behavior.
```
KamilSztandur: I think this comment is unnecessary

lib/src/form/advanced_form_controller.dart
```diff
@@ -25,142 +20,135 @@ class AdvancedFormController
...
-  List<dynamic> _initialFieldsState = const <dynamic>[];
-
-  final Set<AdvancedFieldController<dynamic, dynamic>> _ownedFields = {};
-  final List<VoidCallback> _childCleanups = [];
 
-  /// Takes ownership of registered fields. Will dispose all controllers when
-  /// the form group is disposed.
-  /// Fields are expected to be filled with initial states.
+  /// Takes ownership of the [fields], disposing them when this form is
+  /// disposed. They are expected to be filled with their initial states.
```
KamilSztandur: It feels like this comment doesn't explain the method purpose very well. This method registers certain fields as children of this form + disposes them when it makes sense.

>  They are expected to be filled with their initial states.

It doesn't make sense to mention it here

lib/src/form/advanced_form_controller.dart
```diff
@@ -233,20 +206,55 @@ class AdvancedFormController
...
+    for (final subform in value.subforms) {
+      subform.dispose();
+    }
+    _onValuesChanged.dispose();
+    _onStatusChanged.dispose();
+    super.dispose();
+  }
+
+  Future<bool> _runValidate({required bool enableAutovalidate}) async {
+    // Eager list to prevent short circuits; all fields/subforms must be called.
```
KamilSztandur: Unnecessary comment, I think

lib/src/form/advanced_form_controller.dart
```diff
@@ -233,20 +206,55 @@ class AdvancedFormController
...
+      for (final field in value.fields) field.validate(),
+      for (final subform in value.subforms)
+        subform.validate(enableAutovalidate: enableAutovalidate),
+    ]);
+
+    return results.every((result) => result);
+  }
+
+  // Recurses through the subforms' own methods rather than flattening to
+  // `allFields`, so a subclass that overrides one of them still gets called.
```
KamilSztandur, line 281: Unnecessary comment, I think

lib/src/form/advanced_form_controller.dart
```diff
@@ -273,71 +281,60 @@ class AdvancedFormController
...
   }
 
   void _handleStatusChanged() {
-    final subformsValidating = value.subforms.any(
-      (subform) => subform.value.validating,
-    );
-    final fieldsValidating = value.fields.any(
-      (field) => field.value.isInProgress,
+    // The aggregates are derived on read, so there is nothing to write — but
+    // their readers still have to be told to look again.
```
KamilSztandur: Unnecessary comment, I think

lib/src/form/advanced_form_controller.dart
```diff
@@ -273,71 +281,60 @@ class AdvancedFormController
...
+    notifyListeners();
+    _onStatusChanged.notifyListeners();
+  }
+
+  // The one aggregate that cannot be derived on read: it needs the baseline
+  // captured at `registerFields`.
```
KamilSztandur: Unnecessary comment, I think

lib/src/form/advanced_form_controller.dart
```diff
@@ -273,71 +281,60 @@ class AdvancedFormController
...
-    super.dispose();
+    _value = newValue;
+    notifyListeners();
   }
 }
 
-/// An immutable snapshot of an [AdvancedFormController] — which fields and
-/// subforms it owns, whether the user has changed anything, whether validation
-/// applies, and whether async validation is still running.
+/// A snapshot of an [AdvancedFormController] — which fields and subforms it
```
KamilSztandur: "State" instead of "snapshot" would be better

lib/src/form/advanced_form_controller.dart
```diff
@@ -349,18 +346,33 @@ class AdvancedFormState {
...
-  /// Returns true if fields are currently being validated.
-  final bool validating;
-
-  /// List of this form's fields including subforms' fields.
+  /// This form's fields including every subform's fields.
   Iterable<AdvancedFieldController<dynamic, dynamic>> get allFields =>
       fields.followedBy(subforms.expand((e) => e.value.allFields));
 
-  /// Map of all validation errors (including subforms') grouped by fields.
+  /// Whether an async check is in flight anywhere in the tree.
```
KamilSztandur: Not just async check, but any check at this point

lib/src/form/advanced_form_controller.dart
```diff
@@ -349,18 +346,33 @@ class AdvancedFormState {
...
+  bool get validating => allFields.any((field) => field.value.isInProgress);
+
+  /// Whether some field's async check could not complete. Not sticky: the next
+  /// [AdvancedFormController.validate] retries every failed round.
```
KamilSztandur, line 403: Same, not just async checks

lib/src/form/advanced_form_controller.dart
```diff
@@ -349,18 +346,33 @@ class AdvancedFormState {
...
+  bool get hasFailedValidation =>
+      allFields.any((field) => field.value.isFailedValidation);
+
+  /// Whether every field in the tree is [FieldStatus.valid] right now.
+  ///
+  /// A snapshot of **known** errors, so it is true on a quiet form where
+  /// nothing has been checked yet. Use it to enable a submit button;
+  /// `await AdvancedFormController.validate()` is the guarantee.
```
KamilSztandur: This part of the comment is a bit weird. Just suggest to run `validate()` first

lib/src/form/advanced_form_controller.dart
```diff
@@ -349,18 +346,33 @@ class AdvancedFormState {
...
+  bool get canSubmit => allFields.every((field) => field.value.isValid);
+
+  /// Every error in the tree, keyed by field.
+  ///
+  /// Keyed on [AdvancedFieldState.error], so a field invalid from an async
+  /// check appears too. A failed round with no error code does not.
```
KamilSztandur: Unnecessary part of comment

lib/src/form/advanced_form_controller.dart
```diff
@@ -387,15 +397,13 @@ class AdvancedFormState {
...
+          validationEnabled == other.validationEnabled;
 
   @override
   int get hashCode => Object.hash(
         wasModified,
         Object.hashAll(fields),
         Object.hashAllUnordered(subforms),
         validationEnabled,
-        validating,
       );
```
KamilSztandur: Ditto, I think we should lean towards `equatable` and remove this "Maintainer" comment above it

KamilSztandur (changes requested):

lib/src/field/advanced_field_state.dart
mateusz-pietras: preexisting issue - worth noting but out of scope here

lib/src/field/advanced_field_state.dart
mateusz-pietras: ditto

lib/src/field/advanced_text_field_controller.dart
mateusz-pietras: It previously did not behave correctly. Implemented a custom solution, though its a bit lengthy, what do you think?

lib/src/form/advanced_form_controller.dart
mateusz-pietras: Pkg lower constrain is `3.1.0` while `syncValue` is since 3.10

lib/src/form/advanced_form_controller.dart
mateusz-pietras: ditto, lower constraint

lib/src/field/advanced_text_field_controller.dart
mateusz-pietras: I clarified / simplified the comment but IMO introduces value

lib/src/utils/coalescing_call.dart
mateusz-pietras: Internal is available since 3.35 while floor is 3.10 here. Marking as internal requires adding `meta` as a new dependency

lib/src/form/advanced_form_controller.dart
mateusz-pietras: Linter: `dart[public_member_api_docs](https://dart.dev/diagnostics/public_member_api_docs)`

lib/src/form/advanced_form_controller.dart
mateusz-pietras: ditto

lib/src/form/advanced_form_controller.dart
mateusz-pietras: IMO value added - explains why not to do smth that feels kind of obvious when reading the code

lib/src/utils/coalescing_call.dart
PiotrRogulski: meta is a dependency of Flutter anyways, so I think it's safe to use

lib/src/form/advanced_form_controller.dart
mateusz-pietras: Actually here it makes sense cuz sync validator doesnt need this, it instantly resolves to valid / invalid

lib/src/form/advanced_form_controller.dart
mateusz-pietras: I clarified the behavior, please check again whether to keep or remove

lib/src/form/advanced_form_controller.dart
mateusz-pietras: `run` reserves the slot even for a call that does no work. On a disabled form `validate()` returns `true` immediately — move that inside `run` and it becomes the answer any second caller gets. Two `validate()` calls in the same turn: the first claims the slot, the second joins it and gets `true` without a single validator running.

Measured:

| | 1st call | 2nd call | validation passes |
|---|---|---|---|
| guards outside `run` (current) | `true` | `false` | 1 |
| guards inside `run` | `true` | `true` | 0 |

The early `if` isn't the same check — it stops a joining caller from re-running `setAutovalidate(true)`, which broadcasts to the whole tree.

test/src/field/field_controller_test.dart
```diff
@@ -597,6 +1246,205 @@ void main() {
...
+        reconciled('aa', const TextSelection.collapsed(offset: 0), 'aaa')
+            .selection,
+        const TextSelection.collapsed(offset: 0),
+      );
+      expect(
+        reconciled('aaa', const TextSelection.collapsed(offset: 3), 'aa')
+            .selection,
+        const TextSelection.collapsed(offset: 2),
+      );
+    });
```
PiotrRogulski: What if the caret is in the middle (`a^a`) and you set the text to `aaa`? 🙈

I really like this functionality, though

test/src/field/field_controller_test.dart
PiotrRogulski, line 1: This file is huge; consider splitting

pubspec.yaml
```diff
@@ -1,16 +1,18 @@
...
 homepage: https://github.com/leancodepl/leancode_forms.git
 
 environment:
   sdk: ">=3.1.0 <4.0.0"
-  flutter: ">=3.10.0"
+  flutter: ">=3.13.0"
 
 dependencies:
   collection: ^1.17.1
+  equatable: ">=2.0.0 <3.0.0"
```
PiotrRogulski:
```suggestion
  equatable: ^2.1.0"
```
(2.1.0 and not just 2.0.0 because of `Equatable` itself becoming a mixin class in 2.1.0)

pubspec.yaml
```diff
@@ -1,16 +1,18 @@
...
   flutter:
     sdk: flutter
+  meta: ">=1.9.0 <2.0.0"
```
PiotrRogulski: Also can be a caret

lib/src/field/async_validation.dart
```diff
@@ -0,0 +1,133 @@
...
+  /// [AdvancedFieldController.validate] ignores it — it flushes instead.
+  final Duration debounce;
+
+  /// How long [validator] may run before it is abandoned. Null — the default —
+  /// means no bound, so a [validator] that never settles hangs
+  /// [AdvancedFieldController.validate] forever. On expiry the field moves to
+  /// [FieldStatus.failedValidation] and the abandoned result is dropped.
+  final Duration? timeout;
+
+  /// Called when a round fails. If omitted, the failure goes to
```
PiotrRogulski, line 88: I see many uses of the term `round` used in the comments and docs. I'm guessing it means "a round [of validation]", but a user just looking at any doc comment randomly might get confused 🤔

Also, the use of `round` can suggest that validation can perform multiple rounds before it finishes? Is that the case?

lib/src/field/advanced_field_state.dart
PiotrRogulski: Bump 🤗

lib/src/field/advanced_text_field_controller.dart
KamilSztandur: It still has a bit unnatural language

KamilSztandur (approved):

lib/src/form/advanced_form_controller.dart
PiotrRogulski: 😭

lib/src/form/advanced_form_controller.dart
PiotrRogulski: 😔

lib/src/field/advanced_field_controller.dart
```diff
@@ -67,471 +43,480 @@ class AdvancedFieldController<T, E extends Object>
...
-        validationError: null,
-        asyncError: null,
-        status: FieldStatus.valid,
-      ),
-    );
+  /// Re-runs the **sync** validator if the gate is open, and nothing otherwise.
+  ///
+  /// This is what a dependency's change means for this field: its own value did
+  /// not change, so [AdvancedFieldState.asyncError], the verdict and
+  /// [lastFailure] are left alone and no network call is owed. Read-only fields
```
PiotrRogulski: > no network call is owed

async validation does not necessarily involve networking, though 🤔

lib/src/field/advanced_field_controller.dart
```diff
@@ -67,471 +43,480 @@ class AdvancedFieldController<T, E extends Object>
...
-      );
-    });
+    return _report(await _beginRound(_value.value, debounced: false).done);
+  }
+
+  // A superseded round cannot report success, and neither can a failed one —
+  // `failedValidation` is not `isValid`. Every conjunct is load-bearing:
+  // without `notSuperseded`, a round cancelled into a valid state (say by
+  // `clearErrors`) passes; without `isValid`, an invalid or failed round does;
+  // without `!_isDisposed`, a round that settled just before disposal does.
```
PiotrRogulski: This sentence is very hard to read and understand

lib/src/field/advanced_text_field_controller.dart
```diff
@@ -18,41 +21,140 @@ class AdvancedTextFieldController<E extends Object>
...
+    if (oldText == text) {
+      return;
+    }
+
+    _reconciling = true;
+    textController.value = TextEditingValue(
+      text: text,
+      selection: _mapSelection(textController.selection, oldText, text),
+    );
+    _reconciling = false;
```
PiotrRogulski: The assignment between the two `_reconciling` assignments is sync, so no-one can read the temporary `true` value

lib/src/field/advanced_text_field_controller.dart
```diff
@@ -18,41 +21,140 @@ class AdvancedTextFieldController<E extends Object>
...
+
+    // Neither anchor may reach past the shorter text, so they cannot overlap.
+    final shorter =
+        oldText.length < newText.length ? oldText.length : newText.length;
+    var prefix = 0;
+    while (prefix < shorter && oldText[prefix] == newText[prefix]) {
+      prefix++;
+    }
+    // The scan compares UTF-16 code units, so it can stop halfway through a
+    // character such as an emoji. One unit back is the character boundary.
```
PiotrRogulski, line 95: Would iterating over `.characters` help? (and using character positions instead of codepoint positions)

mateusz-pietras: TextSelection offsets are UTF-16 units, so graphemes are the wrong unit. Head scan already
guarded; tail scan proven unable to split from a valid selection — test added

lib/src/field/advanced_field_controller.dart
mateusz-pietras: true

test/src/field/field_controller_test.dart
mateusz-pietras: Caret stays at 1; test added

mateusz-pietras: It stays at 1. `a|a` -> `aaa` is ambiguous — the new character could have gone in on either side of the caret — so the mapping leaves the caret alone rather than moving it under the user. Pinned now by `a caret between two identical characters stays where it was`.

lib/src/field/async_validation.dart
mateusz-pietras: Fair — it was never defined. `AsyncValidation` now defines it once: one run of the validator over one value is a round. And no, a field runs at most one round at a time; a new value replaces the round in progress instead of queueing a second one. That is in the same doc now.

lib/src/field/advanced_text_field_controller.dart
mateusz-pietras: Removed the flag. Small correction on the reason though: the write to `textController.value` *does* re-enter `_onTextControllerChanged` synchronously. It is harmless for a different reason — `_reconcile` only ever writes `fieldValue` itself, so the texts match by then and the re-entry returns at the first check. Mutation-tested: 89/89 passed without the flag. That invariant is now a comment, plus a test that pins one-pass settling for a value-transforming subclass.

lib/src/field/advanced_text_field_controller.dart
mateusz-pietras: Cannot use character positions here — `TextSelection` offsets are UTF-16 code units, so grapheme indices would be the wrong unit for the API. The head scan already has a split guard (`_splitsCharacter`). The tail scan cannot produce a split from a valid selection: the tail boundary splits a pair in both strings at once, so any offset landing there was already invalid on the way in. Added a test for that half using U+1D400 / U+1D000, which share their low surrogate.

KamilSztandur (approved):

README.md
```diff
@@ -120,11 +120,20 @@ ValueListenableBuilder<AdvancedFormState>(
 );
 ```
 
-`AdvancedFormState` carries `wasModified` (any field changed since `registerFields`), `validating` (an async check is in flight), `fields` / `subforms`, `validationEnabled`, and `validationErrors` — every error in the form tree, keyed by field, for an error summary. With `provider`, `context.select<SignupFormController, bool>((c) => c.value.validating)` subscribes to one slice.
+`AdvancedFormState` carries `wasModified` (any field changed since `registerFields`), `fields` / `subforms`, and `validationEnabled`. Four members are derived from the fields on every read, so they can never outlive the child that justified them:
```
PiotrRogulski, line 123: > Four members are derived from the fields on every read, so they can never outlive the child that justified them:

I'm not sure what this sentence means 🤔

PiotrRogulski (approved):

README.md
KamilSztandur: Yeah I think this sentence can be deleted. I believe it was automatically added after tackling some disposing issues

KamilSztandur: Anyway, it's not that urgent now because there's a separate PR to update the docs :)

PR #54, opened 2026-08-13
feat: rework validation modes

mateusz-pietras:


lib/src/form/child_wiring.dart
```diff
@@ -0,0 +1,99 @@
+part of 'advanced_form_controller.dart';
+
+// Listeners on child fields and subforms, and what to update when they change.
+mixin _ChildWiring on ChangeNotifier {
```
mateusz-pietras, line 4: Is it a good idea of extracting it here? I'm worried that the public API will be harder to lookup - wdyt?

lib/src/field/async_validation.dart
```diff
@@ -133,7 +133,7 @@ void _report(String? field, String action, Object error, StackTrace stack) =>
       FlutterErrorDetails(
         exception: error,
         stack: stack,
-        library: 'advanced_forms',
+        library: 'leancode_forms',
```
KamilSztandur, line 136: Why revert the name :D

lib/src/field/advanced_text_field_controller.dart
```diff
@@ -28,24 +27,6 @@ class AdvancedTextFieldController<E extends Object>
...
-        'Cannot use the focusNode of a disposed AdvancedTextFieldController.',
-      );
-    }
-
-    return _focusNode ??= FocusNode(
-      debugLabel:
-          'AdvancedTextFieldController${name?.isNotEmpty ?? false ? '($name)' : ''}',
-    );
-  }
-
```
KamilSztandur, line 48: I guess it's removed because AdvancedFieldController already provides it, but now I think that we should be able to pass our own focus node. Add it to constructor

lib/src/field/advanced_field_state.dart
```diff
@@ -38,7 +38,7 @@ class AdvancedFieldState<T, E extends Object> with Equatable {
     required this.value,
     this.validationError,
     this.asyncError,
-    this.autovalidate = false,
+    this.mode = ValidationMode.disabled,
```
KamilSztandur: IIRC we wanted to change the name from `disabled` to `manual`

lib/src/field/validation_round.dart
```diff
@@ -0,0 +1,210 @@
...
+
+typedef _Result<E extends Object> = ({
+  E? verdict,
+  AsyncValidationFailure? failure,
+});
+
+// Runs async validation: starts a pass, waits for the result, and writes it
+// only if still current. Round tracking lives on the controller (`_currentRound`,
+// `_hasVerdict`, `_lastFailure`).
+extension _ValidationRounds<T, E extends Object>
```
KamilSztandur: Let's do it as `mixin` if possible, so it's explicitly declared where these methods are coming from when reading `advanced_field_controller.dart` code

lib/src/form/child_wiring.dart
KamilSztandur: As I said in the other comment, as long as it's `mixin`, not `extension` it's findable, then it's fine

KamilSztandur: Splitting this logic into few `mixin`s was a very good idea. Much more readable when logic is grouped

lib/src/form/advanced_form_state.dart
```diff
@@ -0,0 +1,101 @@
+part of 'advanced_form_controller.dart';
```
KamilSztandur: I doesn't have to be `part of` this file

KamilSztandur: It's public API, just export it separately

lib/src/field/advanced_field_controller.dart
KamilSztandur, line 44: Oof I didn't notice it earlier. We have to allow developers to pass their own FocusNodes

lib/src/field/advanced_field_state.dart
```diff
@@ -53,10 +53,9 @@ class AdvancedFieldState<T, E extends Object> with Equatable {
   /// The error the async round recorded. Only the async pipeline writes it.
   final E? asyncError;
 
-  /// Whether a value change runs the validators. With it off,
-  /// [AdvancedFieldController.setValue] stores the value, clears both errors
-  /// and resets the status to [FieldStatus.valid].
-  final bool autovalidate;
+  /// When this field validates itself. The **effective** mode: a field whose
+  /// form has validation switched off reports [ValidationMode.disabled].
+  final ValidationMode mode;
```
PiotrRogulski: I'd name this field `validationMode`

lib/src/field/focus_handling.dart
```diff
@@ -0,0 +1,60 @@
...
+// the [handleUnfocus] the rest of the controller already knows how to act on.
+mixin _FocusHandling on ChangeNotifier {
+  /// Whether this controller has been disposed. Once true it stays true.
+  bool get isDisposed;
+
+  /// Optional label used in reported errors and as the `FocusNode` debug label.
+  /// Not used for identity — fields are identified by reference.
+  String? get name;
+
+  /// Tells the field the user has left it.
```
PiotrRogulski: > the user has left it

IDK, but sounds weird to me

lib/src/form/relations.dart
```diff
@@ -0,0 +1,51 @@
+part of 'advanced_form_controller.dart';
+
+// Listeners on fields this form does not own — for external callbacks only.
+// Does not update the form's own state.
+mixin _Relations on ChangeNotifier {
```
PiotrRogulski, line 5: Even though it's private, maybe we can tighten the `on` type to the controller type, instead of allowing any ChangeNotifier?

The similar thing goes for other mixins introduced in this PR

lib/src/field/advanced_field_controller.dart
```diff
@@ -120,6 +132,46 @@ class AdvancedFieldController<T, E extends Object>
...
+      return;
+    }
+
+    // Unawaited, so a throwing validator would escape into the zone.
+    unawaited(
+      validate().catchError((Object error, StackTrace stackTrace) {
+        _report(name, 'validating after focus loss', error, stackTrace);
+        return false;
+      }),
+    );
```
PiotrRogulski: I'd override `handleUnfocus` as async and use regular await + try/catch

lib/src/form/child_wiring.dart
```diff
@@ -0,0 +1,99 @@
...
+
+  /// Re-runs the **sync** validator on every leaf field the mode allows.
+  void revalidateSync();
+
+  /// Returns this form's own field values, excluding subforms'.
+  List<dynamic> getFieldValues();
+
+  void _setState(AdvancedFormState newValue);
+
+  final _onValuesChanged = ChangeNotifier();
```
PiotrRogulski, line 19: Is this notifier (and `_onStatusChanged`) disposed of properly? 🤔

test/src/field/field_controller_test.dart
PiotrRogulski, line 1: Now at 1763 lines, tl;dr — split up into multiple files mayhaps?

lib/src/field/async_validation.dart
mateusz-pietras: ai artifact I guess, consistent with the rest of the codebase for now. I will replace all in next PR

lib/src/form/child_wiring.dart
mateusz-pietras: yes, in form controller's dispose()

test/src/field/field_validate_test.dart
```diff
@@ -0,0 +1,217 @@
...
+
+    test('reports a superseded round as not known good', () async {
+      final (:field, :validated) = makeAsyncField(
+        validatorDelay: const Duration(milliseconds: 100),
+        mode: ValidationMode.manual,
+      );
+      addTearDown(field.dispose);
+
+      final result = field.validate();
+      await Future<void>.delayed(const Duration(milliseconds: 20));
```
PiotrRogulski, line 115: pumpEventQueue instead of all these near-zero delays? 🤔

PiotrRogulski (approved):

lib/src/field/validation_round.dart
mateusz-pietras: Changed to extension type - mixin involves too much boilerplate in this case

test/src/field/field_validate_test.dart
mateusz-pietras: These delays are strictly tied to the debounce value, I'd name them explicitly here, since debounce is defined here as e.g. 10ms

KamilSztandur (approved):

PR #55, opened 2026-08-13
docs: add AI agent skill for building forms with leancode_forms

mateusz-pietras:
Agent skill at `.claude/skills/leancode-forms/SKILL.md`. Hardened over 5 subagent test→verify→refine iterations; all snippets compile against the package.

### DIFF
--- a/.claude/skills/leancode-forms/SKILL.md
+++ b/.claude/skills/leancode-forms/SKILL.md
@@ -92,12 +92,12 @@ Two rules prevent most bugs. **Call `registerFields()` once, in the constructor,
 
 ## Field controllers
 
-| Controller | Value `T` | Constructor and notes |
+| Controller | Value type | Constructor and notes |
 | --- | --- | --- |
 | `AdvancedTextFieldController<E>` | `String` | `{initialValue = '', validator, asyncValidation, name}`; owns `textController` + `focusNode`, has `focus()` |
 | `AdvancedBooleanFieldController<E>` | `bool` | `{initialValue = false, validator, asyncValidation, name}` |
-| `AdvancedSingleSelectFieldController<V, E>` | `V?` | `{required V? initialValue, required List<V> options, validator, name}`; set with `select(V? option)`, `null` clears; no async validation |
-| `AdvancedMultiSelectFieldController<V, E>` | `Set<V>` | `{required Set<V> initialValue, required List<V> options, validator, name}`; `toggleElement` / `addValue` / `removeValue`; no async validation |
+| `AdvancedSingleSelectFieldController<V, E>` | `V?` | `{required V? initialValue, required List<V> options, validator, asyncValidation, name}`; set with `select(V? option)`, `null` clears |
+| `AdvancedMultiSelectFieldController<V, E>` | `Set<V>` | `{required Set<V> initialValue, required List<V> options, validator, asyncValidation, name}`; set with `toggleElement` / `addValue` / `removeValue` |
 
 `name` labels the field in error reports and as the `FocusNode` debug label. All support `reset()`, `markReadOnly()` / `unmarkReadOnly()`, `setError()`, `clearErrors()`, `setAutovalidate()`, `getValueSetter()` (a `ValueSetter<T>?` — `null` while read-only, so a widget with a nullable `onChanged` disables itself), `validate()`, `subscribeToFields()`.
 
@@ -110,8 +110,8 @@ Two rules prevent most bugs. **Call `registerFields()` once, in the constructor,
 class PhoneFieldController extends AdvancedTextFieldController<MyError> {
   PhoneFieldController({
     String initialValue = '',
-    Validator<String, MyError>? validator, // spell the arity out when forwarding
-  }) : super(initialValue: _digits(initialValue), validator: validator);
+    super.validator, // super-parameters forward the rest unchanged
+  }) : super(initialValue: _digits(initialValue));
 
   static String _digits(String value) => value.replaceAll(RegExp(r'\D'), '');
 
@@ -121,7 +121,7 @@ class PhoneFieldController extends AdvancedTextFieldController<MyError> {
 }
 ```
 
-Why it works: every write to `textController` — keystroke, paste or programmatic `.text =` — goes through the public `setValue`, so the override always runs, and the transformed value is written back into `textController` in the same turn with the caret kept on the same characters. Normalize `initialValue` before it reaches `super`, as above — `reset()` returns to *that* value without passing through `setValue`. A super-parameter (`super.validator`) cannot be combined with an explicit `super(...)` call, so forward by hand: `Validator<String, E>?`, `AsyncValidation<String, E>?`. The subclass may stay generic — `class PhoneFieldController<E extends Object> extends AdvancedTextFieldController<E>` — which is what a reusable field widget wants.
+Why it works: every write to `textController` — keystroke, paste or programmatic `.text =` — goes through the public `setValue`, so the override always runs, and the transformed value is written back into `textController` in the same turn with the caret kept on the same characters. Normalize `initialValue` before it reaches `super`, as above — `reset()` returns to *that* value without passing through `setValue`. Named super-parameters mix freely with the explicit `super(...)` call that carries the normalized value, so forward everything you do not transform as `super.validator`, `super.asyncValidation`, `super.name`. The subclass may stay generic — `class PhoneFieldController<E extends Object> extends AdvancedTextFieldController<E>` — which is what a reusable field widget wants.
 
 ### Field state — `AdvancedFieldState<T, E>`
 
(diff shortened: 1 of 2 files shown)


.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+}
+```
+
+Default UX out of the box: fields stay quiet until the first `validate()` (usually the submit), then give live feedback on every edit.
+
+Two rules prevent most bugs. **Call `registerFields()` once, in the constructor, with every field** (the parameter is `List<AdvancedFieldController<dynamic, dynamic>>`, so mixed value types go in one list) — the form then owns their lifecycle (disposal, `wasModified` tracking, `validate()`, `resetAll()`...), and a second call *replaces* the list, so earlier fields stop participating though they are still disposed with the form. **Bind text widgets to `field.textController`** — the field owns a `TextEditingController` and a `FocusNode` and keeps them in two-way sync, so do not allocate your own and do not wire `onChanged` + `initialValue` for text fields.
+
+## Field controllers
+
+| Controller | Value `T` | Constructor and notes |
```
PiotrRogulski:
```suggestion
| Controller | Value type | Constructor and notes |
```

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+| --- | --- | --- |
+| `AdvancedTextFieldController<E>` | `String` | `{initialValue = '', validator, asyncValidation, name}`; owns `textController` + `focusNode`, has `focus()` |
+| `AdvancedBooleanFieldController<E>` | `bool` | `{initialValue = false, validator, asyncValidation, name}` |
+| `AdvancedSingleSelectFieldController<V, E>` | `V?` | `{required V? initialValue, required List<V> options, validator, name}`; set with `select(V? option)`, `null` clears; no async validation |
+| `AdvancedMultiSelectFieldController<V, E>` | `Set<V>` | `{required Set<V> initialValue, required List<V> options, validator, name}`; `toggleElement` / `addValue` / `removeValue`; no async validation |
```
PiotrRogulski: Hmm, I don't like that we mix `Element` and `Value` in the methods for this field 🤔

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+- Need something else? `AdvancedFieldController<T, E>` is **concrete** — construct it directly for any value with no controller of its own, a slider, a stepper, a rating or a derived total: `AdvancedFieldController<int, MyError>(initialValue: 0, name: 'total')`, whose full constructor is `{required T initialValue, Validator<T, E>? validator, AsyncValidation<T, E>? asyncValidation, String? name}`. It validates like any other field, and ordinary user writes need no `force:` — that is only for writing to a field you have marked read-only. Extend it, forwarding those, only when you want domain methods on top of `void setValue(T newValue, {bool force = false})`. To transform the text the user types (normalize, mask, uppercase), extend `AdvancedTextFieldController<E>` and override `setValue` — the supported way:
+
+```dart
+class PhoneFieldController extends AdvancedTextFieldController<MyError> {
+  PhoneFieldController({
+    String initialValue = '',
+    Validator<String, MyError>? validator, // spell the arity out when forwarding
```
PiotrRogulski: Isn't arity always the same for every validator? 🤔 Also could be `super.validator`

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+
+  static String _digits(String value) => value.replaceAll(RegExp(r'\D'), '');
+
+  @override
+  void setValue(String newValue, {bool force = false}) =>
+      super.setValue(_digits(newValue), force: force);
+}
+```
+
+Why it works: every write to `textController` — keystroke, paste or programmatic `.text =` — goes through the public `setValue`, so the override always runs, and the transformed value is written back into `textController` in the same turn with the caret kept on the same characters. Normalize `initialValue` before it reaches `super`, as above — `reset()` returns to *that* value without passing through `setValue`. A super-parameter (`super.validator`) cannot be combined with an explicit `super(...)` call, so forward by hand: `Validator<String, E>?`, `AsyncValidation<String, E>?`. The subclass may stay generic — `class PhoneFieldController<E extends Object> extends AdvancedTextFieldController<E>` — which is what a reusable field widget wants.
```
PiotrRogulski: > A super-parameter (`super.validator`) cannot be combined with an explicit `super(...)` call

Not true. This is valid:

```dart
class C {
  C({this.x, this.y});
  
  final int? x, y;
}

class D extends C {
  D({super.x}) : super(y: 42);
}
```

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+            DropdownMenuItem(value: option, child: Text(option.name)),
+        ],
+        onChanged: state.readOnly ? null : form.country.select,
+      ),
+    );
+  },
+)
+```
+
+`ErrorTranslator<E>` is `String Function(E)` from the package — write one per error enum: `String translate(MyError e) => switch (e) { MyError.required => 'Required', MyError.tooShort => 'Too short', … };`. It takes a **non-null** `E` while `state.error` is `E?`, so promote before calling it, as above. In a real app wrap each binding once in a reusable widget taking `field` + `ErrorTranslator<E> translateError` and call it everywhere; a generic select wrapper binds `AdvancedFieldBuilder<V?, E>` over `AdvancedSingleSelectFieldController<V, E>`.
```
PiotrRogulski: > write one per error enum

I don't think that's always possible; the same error type may require different labels in different fields

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+validator: filled(MyError.required) &
+    ((String? value) =>
+        (value?.contains('@') ?? false) ? null : MyError.invalidEmail),
+```
+
+Do not wrap the fields in Flutter's `Form` — you need no `Form`, no `GlobalKey<FormState>` and no `autovalidateMode`; this package *is* the validation system, and a second one fights it. (A `TextFormField` outside a `Form` is fine — unregistered, it is a `TextField` with a decoration.)
+
+## Async validation
+
+For checks that live on the server ("is this username taken?"). Only on text and boolean controllers. A verdict your own save call returns is not this — push it with `setError` (see *Server errors*).
```
PiotrRogulski: > Only on text and boolean controllers

Hmm, why?

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+
+**Two fields watching each other** (one rule over a pair, "at least one traveler"): wire them in the **constructor body**, after `registerFields`, and never in a cascade — `adults.subscribeToFields([children]); children.subscribeToFields([adults]);`. A cascade `..subscribeToFields([sibling])` inside a `late final` initializer evaluates the sibling **eagerly**, so a mutual pair **stack-overflows at construction**; a mutually-referencing `late final` pair also needs an **explicit type** (`late final AdvancedTextFieldController<MyError> adults = ...`) or the analyzer reports `top_level_cycle`. No loop results — the subscription re-runs only when a watched **value** changes, not when a sibling's error or status changes. An error only appears on a field whose **own** validator returns it, so the pair's rule goes in **both** validators; the subscriptions are only what re-runs them. `validateAll: true` is the blunt alternative — it replaces the subscriptions, not the duplication.
```
PiotrRogulski: > "at least one traveler"

What does this mean? 👀

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+void enableGift() => addSubform(gift);
+void disableGift() => removeSubform(gift, close: false); // you now own `gift`
+
+@override
+void dispose() {
+  if (!value.subforms.contains(gift)) {
+    gift.dispose(); // the parent only disposes subforms still attached
+  }
+  super.dispose();
+}
```
PiotrRogulski: Is this something that forms have to override on their own? if not, I wouldn't include this snippet here

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,461 @@
...
+  final String label;
+
+  @override
+  Widget build(BuildContext context) {
+    return AdvancedFieldBuilder<String, SignupError>(
+      field: field,
+      builder: (context, state, _) => TextFormField(
+        controller: field.textController, // NEVER your own TextEditingController
+        focusNode: field.focusNode,
+        readOnly: state.readOnly, // add enabled: !state.readOnly to also look disabled
```
KamilSztandur, line 74: Good to know some people may be confused. I don't think we can do anything with that, honestly. These fields as separate flags make sense

KamilSztandur (approved):
It looks good, but take care of Piotr's suggestions :)

.claude/skills/leancode-forms/SKILL.md
mateusz-pietras: Agreed the naming is inconsistent, but `toggleElement` / `addValue` / `removeValue` are the existing public API, carried over from `MultiSelectFieldCubit`. The skill has to document what the package ships. Renaming is a separate breaking change — worth a follow-up issue.

.claude/skills/leancode-forms/SKILL.md
mateusz-pietras: Agreed, the two flags are separate on purpose and the package cannot pick for you. Documented the mapping in the README instead (885f70e): nullable-callback widgets grey out on their own; a text field also needs `enabled: !state.readOnly` when it should look frozen.

mateusz-pietras: Documented

.claude/skills/leancode-forms/SKILL.md
PiotrRogulski: Yes, I mentioned this to maybe make a change in the package itself 😁

.claude/skills/leancode-forms/SKILL.md
```diff
@@ -0,0 +1,453 @@
...
+- Need something else? `AdvancedFieldController<T, E>` is **concrete** — construct it directly for any value with no controller of its own, a slider, a stepper, a rating or a derived total: `AdvancedFieldController<int, MyError>(initialValue: 0, name: 'total')`, whose full constructor is `{required T initialValue, Validator<T, E>? validator, AsyncValidation<T, E>? asyncValidation, String? name}`. It validates like any other field, and ordinary user writes need no `force:` — that is only for writing to a field you have marked read-only. Extend it, forwarding those, only when you want domain methods on top of `void setValue(T newValue, {bool force = false})`. To transform the text the user types (normalize, mask, uppercase), extend `AdvancedTextFieldController<E>` and override `setValue` — the supported way:
+
+```dart
+class PhoneFieldController extends AdvancedTextFieldController<MyError> {
+  PhoneFieldController({
+    String initialValue = '',
+    super.validator, // super-parameters forward the rest unchanged
```
PiotrRogulski, line 113:
```suggestion
    super.validator,
```

PiotrRogulski (approved):

PR #57, opened 2026-08-13
feat(form): add addRelation for cross-field value dependencies

mateusz-pietras:
## What

Adds `AdvancedFormController.addRelation(source, select, onChange)`: calls `onChange` whenever the part of `source`'s value picked by `select` changes.

## Why

0.1.x codebases wired "when field B changes, update field A" with app-side `onValueChange` stream extensions registered through `addDisposable`. 0.2.0 removed both, leaving a manual `addListener` + `removeListener`-in-`dispose()` recipe. `addRelation` restores the one-line call site with automatic cleanup:

```dart
addRelation(
  sourceAccount,
  (account) => account?.commonAccount.currency,
  amount.setCurrency,
);
```

## Design notes

- Lives on the form because the form is the single owner in 0.2.0 — it can guarantee the listener is removed in `dispose()`.
- Parts are compared with `==`; fires on value change only, never on status-only changes. Matches the old `.distinct()` semantics.
- Uses its own cleanup list, not `_childCleanups` — that one is rebuilt on every `registerFields`/`addSubform`/`removeSubform`, which would silently drop relations.
- Cleanups run before owned fields are disposed (`removeListener` on a disposed notifier throws in debug), with a guard for sources disposed earlier by a different owner.
- Throws `StateError` on a disposed form or source, consistent with the other disposed guards.

## Tests

4 tests in `form_controller_test.dart`: fires with the new part, ignores status-only changes, stops on form dispose, disposed guards. Full suite: 201 passing, analyzer clean.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -134,7 +134,7 @@ class PasswordFormController extends AdvancedFormController {
 }
 ```
 
-`subscribeToFields` listens to the given fields and re-runs this field's **sync** validator whenever any of their values change. The async validator is not re-run: this field's own value did not change, so its last answer is still current. It only revalidates — it does not update this field's *value* in response to another field. For a value dependency ("when B changes, set A"), use `addListener` on the source field and `setValue` on the target.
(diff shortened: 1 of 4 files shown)


PiotrRogulski (approved):

KamilSztandur (approved):
This is purely amazing <3

PR #67, opened 2026-08-18
chore: add step example form

mateusz-pietras:


example/lib/screens/step_form.dart
```diff
@@ -0,0 +1,454 @@
...
+  final confirm = ConfirmStepController();
+
+  late final List<AdvancedFormController> steps = [account, address, confirm];
+
+  /// Which step is on screen. Plain `ValueNotifier`s: the package tracks form
+  /// state, never navigation.
+  final currentStep = ValueNotifier(0);
+
+  /// Whether the save call is in flight — see [_SubmitButton].
+  final isSubmitting = ValueNotifier(false);
```
KamilSztandur: 👀

KamilSztandur: Everyone can set its value. Why is it there?

KamilSztandur: And why it's here, shouldn't "isValidating" be enough?

mateusz-pietras: Fair — it was public and writable. It is private state now, exposed read-only:

```dart
bool get isSubmitting => _isSubmitting;
var _isSubmitting = false;
```

Only `submit()` flips it (`_setSubmitting`), the same way `_currentIndex` is only moved by `next()` / `back()`. The screen reads it with `context.watch`, so no notifier is handed out at all any more.

mateusz-pietras: No — they track different things, and both are needed here.

`AdvancedFormState.validating` (and a field's `isValidating`) covers **async validators**: the package knows a check is in flight because it started it. It knows nothing about the save call in `submit()` — that is our own request, after `await validate()` returned true.

So on this screen:
- `state.isValidating` drives the spinner inside the email field while the "is this taken?" check runs,
- `isSubmitting` disables the Submit button while the save request is out.

Dropping `isSubmitting` would leave a double tap sending two requests. It is set before the first `await` in `submit()` for exactly that reason, and `submit()` early-returns if it is already set. The dartdoc on it now says this.

example/lib/widgets/app_dropdown_field.dart
```diff
@@ -29,20 +29,32 @@ class AppDropdownField<T> extends StatelessWidget {
...
-                  value: option,
-                  child: Text(labelBuilder(option)),
-                ),
-            ],
+          // `DropdownButton` in an `InputDecorator`, not
+          // `DropdownButtonFormField`: the selected value is rendered from
+          // [value], so the field controller stays the single source of truth
+          // and a rebuild cannot lose the selection. (It also compiles on every
+          // Flutter version — the form-field variant renamed `value:` to
+          // `initialValue:` in 3.35.)
```
KamilSztandur: Unnecessary comment

example/lib/widgets/app_dropdown_field.dart
```diff
@@ -29,20 +29,32 @@ class AppDropdownField<T> extends StatelessWidget {
...
             decoration: InputDecoration(
               labelText: label,
-              hintText: hint,
               errorText: errorText,
             ),
+            child: DropdownButtonHideUnderline(
+              child: DropdownButton<T>(
+                value: value,
+                isExpanded: true,
+                hint: hint == null ? null : Text(hint!),
```
KamilSztandur: Use pattern matching with switch-case

example/lib/screens/step_form.dart
```diff
@@ -0,0 +1,576 @@
...
+            final AccountStepController step => _AccountStep(step),
+            final AddressStepController step => _AddressStep(step),
+            final InvoiceStepController step => _InvoiceStep(step),
+            final ConfirmStepController step => _ConfirmStep(step),
+            _ => const SizedBox.shrink(),
+          },
+          const _StepControls(),
+        ],
+      ),
+    );
```
KamilSztandur, line 165:
```dart

final body = switch (step) {
  final AccountStepController step => _AccountStep(step),
  final AddressStepController step => _AddressStep(step),
  final InvoiceStepController step => _InvoiceStep(step),
  final ConfirmStepController step => _ConfirmStep(step),
  _ => null,
};

return SingleChildScrollView(
      child: Column(
        children: [
          ?body, // or if (body != null) body, but Dart 3.3 should have this syntax
          const _StepControls(),
        ],
      ),
    );
```

example/lib/screens/step_form.dart
```diff
@@ -0,0 +1,576 @@
...
+  }
+
+  @override
+  String get title => 'Account';
+
+  final email = AdvancedTextFieldController(
+    // Both sides of `&` must take the same type, so the closure is written
+    // over `String?` like the built-in string validators.
+    validator: filled(ValidationError.empty) &
+        ((String? value) => (value?.contains('@') ?? false)
```
KamilSztandur: Let's extract this to named validator. Right now it looks very ugly

KamilSztandur: Also, this would give us nice example on how to write custom validators

example/lib/widgets/app_dropdown_field.dart
mateusz-pietras: Removed.

example/lib/widgets/app_dropdown_field.dart
mateusz-pietras: Done:

```dart
hint: switch (hint) {
  final hint? => Text(hint),
  null => null,
},
```

example/lib/screens/step_form.dart
mateusz-pietras: Applied, with one change: `?body` needs null-aware elements, which is Dart 3.10 — this package declares `sdk: ">=3.3.0"` and the example `">=3.1.0"`, so it does not compile here. Went with your fallback:

```dart
final body = switch (step) {
  final AccountStepController step => _AccountStep(step),
  final AddressStepController step => _AddressStep(step),
  final InvoiceStepController step => _InvoiceStep(step),
  final ConfirmStepController step => _ConfirmStep(step),
  _ => null,
};

return SingleChildScrollView(
  child: Column(
    children: [
      if (body != null) body,
      const _StepControls(),
    ],
  ),
);
```

Better than the old `_ => const SizedBox.shrink()` either way — no empty widget in the tree.

example/lib/screens/step_form.dart
mateusz-pietras: Agreed, and good call on it doubling as the custom-validator example. Extracted:

```dart
/// A validator of your own is just a function of the value. Write it over
/// `String?`, like the built-in string validators, so it combines with them
/// through `&` and `|` — both sides have to take the same type.
Validator<String?, E> isEmail<E extends Object>(E message) =>
    (value) => value != null && value.contains('@') ? null : message;
```

The call site is now:

```dart
validator: filled(ValidationError.empty) & isEmail(ValidationError.invalidEmail),
```

Kept it generic in `E` so it reads like the built-ins, and the comment that used to explain the `String?` cast now lives on the validator where it belongs.

KamilSztandur (approved):

PR #74, opened 2026-09-01
Fix AdvancedFormState.allFields throwing when a subform's fields are explicitly typed

KamilSztandur:
`AdvancedFormState.allFields` threw a `TypeError` instead of returning the fields
whenever a form had a subform and either field list had been registered through an
explicitly typed literal:

```dart
registerFields(<AdvancedFieldController<dynamic, dynamic>>[email, password]);
```

```
type 'ExpandIterable<AdvancedFormController, AdvancedFieldController<dynamic, dynamic>>'
is not a subtype of type 'Iterable<AdvancedFieldController<dynamic, Object>>' of 'other'
  package:advanced_forms/src/form/advanced_form_state.dart 46:14  AdvancedFormState.allFields
```

`AdvancedFieldController` declares `E extends Object`, so `dynamic` in the `E` position
reifies to its bound and the list's runtime type is
`List<AdvancedFieldController<dynamic, Object>>`. `followedBy` checks at runtime that
`other` carries the same element type, and `subforms.expand(...)` produces
`Iterable<AdvancedFieldController<dynamic, dynamic>>`. The two do not match.

## Why it matters

`allFields` is what you walk to focus the first invalid field after a failed submit, or
to apply a server's per-field errors. The crash therefore landed in the "submit did not
go through" path — exactly when the user is already having a bad time. The condition is
narrow but not exotic: a subform, plus one read of `allFields`.

## The fix

Yield the fields instead of concatenating them; `sync*` does no such check.
`_countedFields`, sitting immediately below in the same file and doing the same walk, was
already written this way, so `allFields` was the odd one out rather than a deliberate
choice.

## Tests

Two added to `test/src/form/form_controller_test.dart`: one pinning the documented order
(own fields, then each subform's, recursively), one reproducing the crash. The second
fails without the fix.

No API change.

### DIFF
--- a/lib/src/form/advanced_form_state.dart
+++ b/lib/src/form/advanced_form_state.dart
@@ -42,8 +42,12 @@ class AdvancedFormState with Equatable {
   final ValidationMode validationMode;
 
   /// This form's fields including every subform's fields.
-  Iterable<AdvancedFieldController<dynamic, dynamic>> get allFields =>
-      fields.followedBy(subforms.expand((e) => e.value.allFields));
+  Iterable<AdvancedFieldController<dynamic, dynamic>> get allFields sync* {
+    yield* fields;
+    for (final subform in subforms) {
+      yield* subform.value.allFields;
+    }
+  }
 
   // Fields that count — disabled subtrees excluded, matching what validate()
   // checks, so validate() and canSubmit always agree.
--- a/lib/src/form/advanced_form_state.dart
+++ b/lib/src/form/advanced_form_state.dart
@@ -42,6 +42,8 @@ class AdvancedFormState with Equatable {
   final ValidationMode validationMode;
 
   /// This form's fields including every subform's fields.
+  // `followedBy` compares element types at runtime and throws when a subform's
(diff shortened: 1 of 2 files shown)


lib/src/form/advanced_form_state.dart
```diff
@@ -42,8 +42,12 @@ class AdvancedFormState with Equatable {
   final ValidationMode validationMode;
 
   /// This form's fields including every subform's fields.
-  Iterable<AdvancedFieldController<dynamic, dynamic>> get allFields =>
-      fields.followedBy(subforms.expand((e) => e.value.allFields));
+  Iterable<AdvancedFieldController<dynamic, dynamic>> get allFields sync* {
```
mateusz-pietras, line 47: I'd add a small note here, explaining why we use this syntax instead of standard operations on Iterable

mateusz-pietras (approved):

lib/src/form/advanced_form_state.dart
PiotrRogulski: Maybe the previous syntax would work with a `.cast()` on the inner collection? 🤔

PR #78, opened 2026-09-08
Keep the page title when a Flutter island starts

KamilSztandur:
The browser tab showed the site's address instead of the page title — on the landing page and on every docs page with a live example.

**Cause:** every island is a `MaterialApp`, and `MaterialApp` renders a `Title` widget with an empty title. The Flutter web engine forwards that to `document.title`, so the moment a demo attaches, the title becomes `""` and browsers fall back to the URL.

**Fix:** both runtimes (`docs_app/lib/flutter-examples/runtime.ts`, `landing/web/landing.js`) install a small guard on `document.title` before the engine loads: writes of an empty string are ignored, everything else goes through, so the titles Next sets on client-side navigation still work.

Verified in headless Chrome:

| Page | Before island | After island |
| --- | --- | --- |
| `/` | advanced_forms — Form validation and state for Flutter | unchanged |
| `/docs/first-form` | Your first form — advanced_forms | unchanged |
| client-side nav → `/docs/installation` → `/docs/focus` (with island) | | Installation — advanced_forms → Focus — advanced_forms |

### DIFF
--- a/docs_app/flutter/lib/support/island_frame.dart
+++ b/docs_app/flutter/lib/support/island_frame.dart
@@ -55,11 +55,8 @@ class _IslandFrameState extends State<IslandFrame> {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
-      // Explicitly null, not the default '': with a title, MaterialApp mounts a
-      // Title widget and the web engine writes that title — here an empty
-      // string — into document.title, and the browser tab shows the address
-      // instead of the page's title. With null, WidgetsApp skips Title on the
-      // web because the host page owns the title (flutter/flutter#152003).
+      // null (not the default ''), or the engine overwrites the page's
+      // document.title with it. See flutter/flutter#152003.
       title: null,
       debugShowCheckedModeBanner: false,
       theme: islandTheme(widget.brightness),
--- a/docs_app/flutter/lib/support/island_frame.dart
+++ b/docs_app/flutter/lib/support/island_frame.dart
@@ -55,6 +55,12 @@ class _IslandFrameState extends State<IslandFrame> {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
+      // Explicitly null, not the default '': with a title, MaterialApp mounts a
+      // Title widget and the web engine writes that title — here an empty
+      // string — into document.title, and the browser tab shows the address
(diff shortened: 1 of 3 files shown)


docs_app/lib/flutter-examples/runtime.ts
```diff
@@ -57,10 +57,33 @@ declare global {
...
+  Object.defineProperty(document, "title", {
+    configurable: true,
+    get() {
+      return get.call(this)
+    },
+    set(value: string) {
+      if (String(value).trim() !== "") set.call(this, value)
+    },
+  })
+}
```
PiotrRogulski: Is this seriously the best way to do this? Can't we prevent the flutter islands from doing this?

KamilSztandur: Fair — there is a proper way. `MaterialApp`'s `title` defaults to `''`, and any non-null title mounts a `Title` widget the engine mirrors into `document.title`; with `title: null` WidgetsApp skips `Title` on the web because the host page owns the title (flutter/flutter#152003). Done in 3e9a718 in `island_frame.dart`, the JS guards are gone. Verified the same way: titles intact on /, /docs/first-form and across client-side navigation.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

docs_app/flutter/lib/support/island_frame.dart
```diff
@@ -55,6 +55,12 @@ class _IslandFrameState extends State<IslandFrame> {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
+      // Explicitly null, not the default '': with a title, MaterialApp mounts a
+      // Title widget and the web engine writes that title — here an empty
+      // string — into document.title, and the browser tab shows the address
+      // instead of the page's title. With null, WidgetsApp skips Title on the
+      // web because the host page owns the title (flutter/flutter#152003).
```
PiotrRogulski: I ain't reading allat

KamilSztandur: Cut to two lines in 36c3333.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PiotrRogulski (approved):

PR #79, opened 2026-09-09
Landing: fix the Lighthouse findings

PiotrRogulski:
## What

Fixes the findings from the desktop and mobile Lighthouse reports of advanced-forms.leancode.co, following the same pass done on [ciach](https://github.com/leancodepl/ciach).

### Performance

- **Fonts self-hosted** in `landing/web/fonts/` (the Latin and Latin-ext variable subsets ciach ships, with the OFL texts). Removes the render-blocking Google Fonts stylesheet and the two preconnects, which were also the only third-party requests and the source of the console error in the report. The Latin subsets are preloaded; the docs app caches `/fonts` for a day.
- **Stylesheet inlined** into `<head>` at build, minified, so nothing render-blocking is left. `landing.css` is no longer copied to `public/`.
- **Flutter engine** starts after `load` on an idle frame instead of as soon as `landing.js` runs, so the 3+ MB bundle no longer competes with the first paint.
- **Chromium CanvasKit** is kept in the bundle again and the bootstrap uses `canvasKitVariant: "auto"`, so Chrome fetches the smaller variant (~a third less over the wire) rather than the full one.
- **Logo images** carry their intrinsic `width`/`height`.
- The dangling `//# sourceMappingURL=flutter.js.map` line is stripped from `flutter_bootstrap.js` (it was a 403 in the report).

### Accessibility

- The source tabs were a `role="tablist"` whose children were labels and a copy button. They are now a `radiogroup` of the radio inputs and their labels, with the copy button beside the strip. Keyboard focus is shown on the focused tab.
- Muted text and code comments were at 4.41:1 and 3.97:1. Dark values are now ciach's (`#84848d`, `#767eaa`); the light-theme muted, comment and success colours were darkened to pass as well.
- Footer links inside running text are underlined instead of relying on colour.

### Docs workflow: preview deployments for pull requests

Previews used to come only from pushes to a dedicated `docs` branch, so a PR like this one never got one. Now a pull request from this repository that touches the docs deploys a Vercel preview and gets its URL in one comment, updated on every push. Only a push to `main` is production; a fork's PR has no secrets and runs the build job alone. The `docs` branch trigger is gone. Runs are serialized per ref, with a newer push to a PR superseding the preview in flight.

## Verification

Jaspr build plus Lighthouse 13 against the built page locally (Flutter bootstrap stubbed, since the bundle needs the Flutter SDK):

| Category | Before (mobile / desktop) | Local after |
| --- | --- | --- |
| Accessibility | 0.89 / 0.89 | 1.00 |
| Best practices | 0.96 / 0.96 | 1.00 |
| SEO | 1.00 | 1.00 |
| Render-blocking savings | 740 ms / 220 ms | none |

`dart analyze`, `dart format`, eslint and prettier pass on the changed files. Screenshots checked the tab strip, focus ring and footer in both themes and on a narrow viewport. The workflow change is exercised by this PR itself: the preview comment below is its output.

## Not addressed

- The remaining mobile performance cost is the Flutter bundle itself (unused JS in `main.dart.js`, CanvasKit size). A renderer or build change is a separate decision.
- `main.dart.js` still has no source map; shipping one for a release build is not worth the deploy size.
- `docs_app/app/global.css` keeps the old `--af-muted` value, so the docs pages would show the same contrast finding if audited.

https://claude.ai/code/session_01TazpPghXzy1LwT77FpxgHJ


landing/web/landing.css
```diff
@@ -1058,11 +1110,19 @@ a.af-pill-accent:hover {
...
 
-.af-example-tabs {
+.af-example-toolbar {
   display: flex;
   align-items: center;
   gap: 0.25rem;
   padding: 0.35rem 0.35rem 0;
+}
+
+.af-example-tabs {
```
Komoszek, line 1120: I would extract all those styles next to the approperiate components

Komoszek (approved):

landing/web/landing.css
PiotrRogulski: Yes, that is something I want to prompt as well (I just kept this PR to the lighthouse stuff only)

PiotrRogulski: Agreed, that's the right shape: the same `@css`-next-to-the-component layout ciach's website uses, with tokens and the reset in a `styles.dart`. It's the whole 1,350-line sheet moving into the Dart styles API, so it goes in a follow-up PR rather than on top of the Lighthouse changes here. Jaspr's style bundling inlines the collected rules into the page, so the follow-up also retires the `_Stylesheet` inlining step this PR adds.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #87, opened 2026-09-20
Let a prefilled field react to the fields it depends on

KamilSztandur:
## Problem

A field the user has never touched never validates on its own. That is on purpose: a form filled from the server should not open covered in errors.

But it also means `subscribeToFields` does nothing for such a field. Example: the instructor is prefilled from the server, the user switches the aircraft, the instructor is no longer allowed to fly it. Nothing happens, because the user never edited the instructor field. Today the only way around it is to write the error by hand with `setError`, which duplicates the validator. Someone migrating two real forms ran into exactly this.

## Solution

`field.markInteracted()`. Call it on a prefilled field and from then on it behaves like a field the user has edited: dependency changes reach it and its sync validator runs. It does not change the value, does not start async validation, and under `manual` mode it still validates nothing. `reset()` makes the field untouched again. There is also a `hasInteracted` getter to read the flag.

The rule itself stays. This only adds a way to opt out of it for one field.

## In this PR

- `markInteracted()` and `hasInteracted` on `AdvancedFieldController`.
- Tests for the new method: untouched field ignores its dependency, reacts after the call, no async round, quiet in manual, reset reverts, disposed throws.
- Docs (validation modes, cross-field, FAQ), SKILL.md and a changelog entry.

`flutter analyze` clean, all 266 tests pass.


KamilSztandur:
Closed: this fixed the wrong side of the migration report. The validation gate stays as it is; a documented workaround follows in a new PR.

PR #88, opened 2026-09-20
Document how to build a select whose options change

KamilSztandur:
## Problem

The built-in select has a fixed list of options. There is no way to change it after the field is created. Common case: the list of instructors depends on the chosen aircraft, or the list of cities loads once the country is known. People hit this and read it as a missing feature.

We do not want to change the package here. A select over a fixed list is the model, and making options part of the field state is a bigger decision. What was missing is a documented way to do it yourself.

## Solution

Docs only, no code change. A new section in the custom-controllers page shows a complete controller built on `AdvancedFieldController<V?, E>` with its own mutable options list and a `setOptions` method. When the selected value is no longer on the new list, it is cleared without counting as a user edit, so an untouched field shows no error until submit, and a field the user had picked from shows "required" right away. The section also warns that membership is checked with `==`, so an option deserialized from the server is not the same object as the one on the list.

## In this PR

- New section "A select whose options change" in `docs/fields/custom.mdx`, with the full controller and the `==` warning.
- A pointer from the single-select docs, a new FAQ entry, and the same controller in SKILL.md.

The controller from the docs was run against the package in a throwaway test before writing it up.


KamilSztandur:
Closed: the changing options list is the real gap and gets fixed in the package instead. New PR follows.

PR #91, opened 2026-09-20
A conditional section in one line

KamilSztandur:
## Problem

A section that applies only sometimes, like invoice details behind an "I need an invoice" switch, takes a relation and two calls in every form:

```dart
addSubform(invoice);
addRelation(needsInvoice, (on) => on, (on) => on ? addSubform(invoice) : removeSubform(invoice));
if (!needsInvoice.fieldValue) removeSubform(invoice); // relations fire on change only
```

The seeding line is easy to forget, and the whole thing says nothing about what it is for. Piotr suggested that `addSubform` should just take the condition and do the attaching and detaching itself.

## Solution

```dart
addSubform(invoice,  enabled: () => needsInvoice.fieldValue);
addSubform(shipping, enabled: () => !sameAsBilling.fieldValue);
addSubform(company,  enabled: () => type.fieldValue == CustomerType.company);
```

One optional parameter on a method everyone already uses. The form evaluates the closure when the section is added and again after every value change anywhere in its tree. While it is true the section is attached. When it turns false the section is detached, exactly as `removeSubform` would do: its fields leave `validate()`, `canSubmit`, `getFieldValues()` and `allFields`, so a payload built from the form has no invoice in it. The form owns the section either way, so its values stay in its fields and are back when the switch flips, and it is disposed with the form. A reset of the checkbox takes the section with it on its own. `removeSubform` drops the condition.

Cost: one closure call and a bool comparison per conditional section on each value change, next to the pass over all fields the form already does for `wasModified`.

Limit, written into the docs: a condition on something outside the form is not re-evaluated for that change. Attaching by hand still covers it.

## In this PR

- `enabled` on `addSubform`, re-evaluated from the existing value-changed handler; attach and detach share the code with `addSubform` and `removeSubform`.
- Tests: detached but owned, attached at once when true, follows the fields it reads, values survive a detach, quiet when the result does not change, reacts to a change inside another subform, out of `validate()` while detached, parent mode reaches a late attach, `resetAll()` re-syncs, `removeSubform` drops the condition, re-attach and replace cases, outside field not followed.
- Example app: the wizard's invoice step uses it; the navigation follows the attached steps.
- Docs (subform patterns, subform index, wizard, relations table), SKILL.md and a changelog entry. The switch-validation-off pattern stays documented as the alternative for when `resetAll` and friends must keep reaching a hidden section.

`flutter analyze` clean in the package and the example, all 273 tests pass.


mateusz-pietras:
IDK, as a user I think I would feel lost in this, when to use removeSubform, when to make it conditional... What about a separate method for that, leaving `add/remove` as an advanced feature, while idk `attachSubform(form, condition)` would stay a default way? I dont like either of these options tbh

KamilSztandur:
@mateusz-pietras After reconsideration I agree that's API doubling, but I think it's in good faith.

Adding subform with "enabled by" hook, then removing it with "removeSubform" will simply deactivate that hook. 

To make some order of this "chaos" I expanded Dart docs for this "enabledBy" parameter, explaining that it's convenient shortcut for most simple use cases

Let me know what you think:)

docs/subforms/patterns.mdx
```diff
@@ -9,28 +9,54 @@ AI-Provenance:
...
 ```
 
+The closure is evaluated at `addSubform` and again whenever a value anywhere in the form changes. While it is true the
+section is attached; when it turns false the section is detached, exactly as `removeSubform` would: its fields leave
+`validate()`, `canSubmit`, `getFieldValues()` and `allFields`, so the payload you build from the form has no invoice
+in it. The form owns the section either way, so **the values stay in its fields** and are there again when the
+checkbox goes back on — and the section is disposed with the form.
+
+Any expression works: `() => type.fieldValue == CustomerType.company`, `() => !sameAsBilling.fieldValue`, a condition
+over two fields. Because the form re-evaluates it after `resetAll()` too, a reset checkbox takes the section with it.
```
PiotrRogulski, line 31: Hmm, this first sentence sounds weird to me (I'd just say that it can be any boolean expression using the fields' values. These examples don't really carry too much value, in my opinion)

lib/src/form/advanced_form_controller.dart
```diff
@@ -227,8 +273,33 @@ class AdvancedFormController
...
+    _runChildCleanups();
+    _setState(value.copyWith(subforms: {...value.subforms}..remove(form)));
+    _wireChildren();
+    _recomputeWasModified();
+  }
+
+  @override
+  void _applySubformConditions() {
+    for (final entry in _subformConditions.entries) {
+      final condition = entry.value;
```
PiotrRogulski, line 286: Destructure the entry into key & value in the loop header

lib/src/form/advanced_form_controller.dart
```diff
@@ -358,3 +427,10 @@ class AdvancedFormController
     notifyListeners();
   }
 }
+
+class _SubformCondition {
+  _SubformCondition(this.enabled, this.last);
+
+  final bool Function() enabled;
+  bool last;
+}
```
PiotrRogulski, line 436: Could be a typedef of a record?

skills/advanced_forms-build-forms/SKILL.md
```diff
@@ -760,13 +755,41 @@ class CheckoutFormController extends AdvancedFormController {
 }
 ```
 
+The closure runs at `addSubform` and again after every value change anywhere in this form's
+tree. True → the section is attached; false → detached, exactly like `removeSubform`: out of
+`validate`, `canSubmit`, `getFieldValues`, `allFields` and every broadcast. **The form owns
+the section either way**, so its values stay in its fields and are back when the switch flips,
+and it is disposed with the form. Any expression goes — `() => type.fieldValue ==
+CustomerType.company`, a condition over two fields. `resetAll()` re-triggers it, so a reset
```
PiotrRogulski, line 763: Ditto

PiotrRogulski (approved):

mateusz-pietras (approved):
LGTM

